PR #9, opened 2016-06-15
Fixed height issue preventing repl from being usable

amilajack:


amilajack:
Before:
screen shot 2016-06-14 at 11 08 17 pm

After:
screen shot 2016-06-14 at 11 10 47 pm

developit:
Ahh great catch, this seems to be an issue in Chrome M51+.  I'll probably swap `100vh` out for `100%` though, just for better browser compatibility.

amilajack:
For the most part, support is pretty good. Don't think any dev's still use IE
screen shot 2016-06-15 at 8 26 50 am

developit:
Yup, I'm using `vh` for the header search. Just in this case the offset parent is already the size of the viewport.

developit:
By the way this is already live! Thanks!

amilajack:
Just saw it. Thanks!

Also I have a lot of free time on my hands this summer. I'm willing to make significant PR contributions to preact. Do you have any suggestions on what I should start working on?

developit:
That would be amazing. The biggest thing we need now is [more tests](https://github.com/developit/preact/tree/master/test/browser), if you're interested in writing them!  Secondly we need more close integrations with other frameworks (we have [preact-redux](https://git.io/preact-redux), but there are so many others like MobX).

PR #15, opened 2016-07-08
Remove setState not being async

timdorr:
Basic batching was folded into React a while ago, so `setState` is now async and queues up changes for the next batch run. One less API difference!

### DIFF
--- a/content/guide/differences-to-react.md
+++ b/content/guide/differences-to-react.md
@@ -61,10 +61,9 @@ Preact actually adds a few convenient features inspired by work in the React com
 
 ## What's Different?
 
-There are currently a couple differences between Preact and React that are more subtle:
+Preact and React have some more subtle differences:
 
 - `render()` accepts a third argument, which is the root node to _replace_, otherwise it appends. This may change slightly in a future version, perhaps auto-detecting that a replacement render is appropriate by inspecting the root node.
-- `setState()` is asynchronous because of batching. This is intentional and does not affect the outward-facing API.
 
 
 [Project Goals]: /about/project-goals


developit:
Good catch. Is react's batching async outside the context of React though? I thought it was only batched within a single outer call.

timdorr:
I believe they batch everything within the same update cycle. [The `setState` code](https://github.com/facebook/react/blob/d1256825bcab63f16d9335a2841558379e7cbddd/src/isomorphic/modern/class/ReactComponent.js#L60-L72) enqueues the state change, which [enqueues an update](https://github.com/facebook/react/blob/25f9f4563edb0b0daa4687da0c75b3ae010e9816/src/renderers/shared/stack/reconciler/ReactUpdateQueue.js#L246).

But regardless of all that, `setState` returns asynchronously. @mweststrate had a good example of the impact of this here: https://medium.com/@mweststrate/3-reasons-why-i-stopped-using-react-setstate-ab73fc67a42e#3a33

developit:
Cool, yeah this looks like roughly the same semantics.

developit:
Once re-render detection lands we can basically remove this section 🙂

jmrog:
This is a pretty old conversation (found it by googling), and I'm admittedly being a bit lazy here (I could go through the Preact code, or write some tests/fiddles, and find out on my own), but I'm a little confused by this change and its implications. Is it the case that, in Preact, `setState` _always_ asynchronously updates component state? This did not used to be the case in React, and it seems like it is still not the case in the latest version (v15.5.4). Sometimes, in React, component state is updated synchronously when `setState` is called. Here is a blog post about this (involving React 13): https://www.bennadel.com/blog/2893-setstate-state-mutation-operation-may-be-synchronous-in-reactjs.htm. And here is a (very sloppily/hastily-written) fiddle that shows the same thing for React v15.5.4: https://jsfiddle.net/8ncaev16/4/

I'm asking primarily because I was in the process of converting a component from React to Preact, and it seemed that I did have to change the behavior in the Preact version of the component. In the React version, `setState` was updating state immediately and forcing a synchronous render of the component, whereas, in the Preact version, this was not so, and I had to call `forceUpdate` to get the same behavior. Adding some logs confirmed this (note the ordering below).

With React, I was reliably getting logs like this:

```
setting state with value true
rerendering component with value true
setState has returned
```

With Preact, without using `forceUpdate`, I was reliably getting logs like this:

```
setting state with value true
setState has returned
rerendering component with value true
```

jmrog:
I decided to be a bit less lazy, and went looking through the source code to get some clarification about this. I now see that rendering is [intentionally asynchronous](https://github.com/developit/preact/blob/52bb91e17a56e6f3dba050c3ae001292b6ff2285/src/render-queue.js#L10) (a different issue from whether `setState` itself is asynchronous, admittedly), which explains the behavior I was seeing in my last comment. I also see that this is documented on the Preact website (under the "What's Added?" section [here](https://preactjs.com/guide/differences-to-react)), so I guess it was just some bad luck that I landed _here_ rather than _there_ when I was googling about the cause of the discrepancy I mentioned. Oh well; that's good enough for me!

developit:
@jmrog I should add - the difference widened a bit (unintentionally) in Preact 8 because we switched to setTimeout() for async.  I've actually just reverted that change and we're back to using Promises for async.  Should be a release soon.

PR #17, opened 2016-07-19
Remove preact-svg, add preact-hyperscript

queckezz:


developit:
Awesome, thanks again!

PR #20, opened 2016-07-21
Remove .preventDefault() in checkbox example

MozMorris:
See https://github.com/developit/preact/issues/241


developit:
Ignore snyk, thanks for doing this!

PR #21, opened 2016-07-22
Fixed pragma setting in getting-started to directly use default's import

tobiasweibel:


developit:
@tobiasweibel I think we'd be better off reversing this - using `h` in the example, with a note below indicating that you can use `preact.h` if you're relying on `preact` as a global.  In terms of recommended usage, using `h()` directly results in smaller builds ;)

tobiasweibel:
OK, created #28 and basically added a sample for default import as well.

PR #26, opened 2016-08-03
fixed logo ellipse stroke gaps

robbiewxyz:
In order for the ellipse animation's stroke-dasharray trick to work seamlessly, the dasharray needs to sum to exactly the circumference of your ellipse.  Here's a close-up GIF of the problem:

Here's a better gif to show the stroke glitches

you can see the strokes glitching where the gaps will close pretty tight before the other end catches up.

Anyway, I did a little math and with the help of [this Wolfram search](http://www.wolframalpha.com/input/?i=ellipse+a%3D196+b%3D75+circumference) and I've got it spinning seamlessly.

Hope it's :smile:!

### DIFF
--- a/src/components/logo.js
+++ b/src/components/logo.js
@@ -26,8 +26,10 @@ export default class Logo extends Component {
 	}
 
 	renderEllipse(fg, deg, offset) {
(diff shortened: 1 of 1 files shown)


developit:
You're a wizard!

PR #29, opened 2016-08-26
Add switch guide

developit:
Thanks @vvo for the idea!


vvo:
LGTM, this is very extensive. You are very fast :)

I will tweet this like a beast once released :D

PR #35, opened 2016-10-27
Update link from preact-shallow-compare -> shallow-compare

tkh44:
I made preact-shallow-compare -> shallow-compare because it works for multiple react like libs. Updated the name and link.


developit:
Nice!

PR #36, opened 2016-10-28
finish first chapter

lcxfs1991:
1. I create a `zh` folder under `content`
2. Please modify webpack config file to enable multiple compiler in order to build pages for different languages.


developit:
Amazing! 😍
I think we might be able to add a language picker or even automatically fetch `zh` if the browser reports that as the selected language. Will post back in a bit when I've found a way.

PR #39, opened 2016-10-31
Address preact #377

pl12133:
In order to clarify the difference between React's SyntheticEvent system and Preact's event system, add footnote to "Synthetic Events" to explain Preact's usage of `addEventListener`.


developit:
Looks good! We can add the specific bits about `onInput` and whatnot later.

PR #46, opened 2016-11-02
move the md to right place

lcxfs1991:
1. there is a file at the wrong place. correct it.
2. a question: why the chinese version is the default now?


developit:
Ooh good catch! sorry I didn't notice the path there. 
Chinese is the default for you because the language picker at the bottom remembers your choice using `localStorage` for convenience 😇

PR #48, opened 2016-11-03
Do you mean kilobytes instead of kilobits?

hrldcpr:
I'm not sure because I can't find the minified gzipped file you're referring to, but just in case you mean kilobytes (kB) not kilobits (kb), here's a patch to fix it.


developit:
Good catch! Definitely meant kilobytes.

PR #51, opened 2016-11-18
Update index.md

ericdfields:
Without yet reading the docs, this part of the sentence read as a statement without a clause to conclude it. I believe you mean to say something along the lines of "Preact retains a large amount of compatibility with React, (such as|but especially|etc) the stateless functional components…"


developit:
Ah - couldn't agree more, and I have found people tend to see the stateless bit and not the classes bit, so they think Preact only supports stateless functions (making it mostly useless)

PR #52, opened 2016-11-19
Improve user language detection

a-xin:
When a user configures the following languages in their browser:
`navigator.languages === ["en-US", "en", "de-DE", "de", "zh-TW", "zh"]`
the page currently renders in Chinese, although English should take precedence.

This PR changes that behaviour and makes sure the order of languages is respected by including English in the detection (instead of only falling back to English if no matching localisation is found).

### DIFF
--- a/src/lib/default-language.js
+++ b/src/lib/default-language.js
@@ -4,6 +4,8 @@ export default function getDefaultLanguage(available={}) {
 		if (langs[i]) {
 			let lang = String(langs[i]).toLowerCase();
 			if (available[lang]) return lang;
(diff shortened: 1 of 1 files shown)


a-xin:
Hey @developit - a quick-fix to make sure webpage renders in the right language.

Another solution that is more involved would be: move the English docs to its own language folder, add `en` to the config, configure `en` as fallback language. If that makes sense, might give that a try in a follow-up PR

developit:
Hi there - I had initially move the English docs under en, but I think it's worth special-casing them since those are written first.

PR #56, opened 2016-11-21
Keep the footer at the bottom

ForsakenHarmony:


developit:
I haven't forgotten this! Just wanted to test it out locally first :)

src/style/index.less
```diff
@@ -93,4 +93,11 @@ content-region {
 	& > main {
 		padding-top: 56px;
 	}
+	
+	display: flex;
+	flex-direction: column;
+}
+
+main {
```
developit: I tested and the only thing I noticed was that I think this block needs to be moved to `L93` and the selector changed to:

```css
& > main, & > footer {
```

developit:
I ended up just making `<main>` have `min-height:95%`.

PR #61, opened 2016-12-02
Update Edit page link to support i18n

montogeek:
This PR updates Edit page link to support the new languages, just one step easier for new contributors.

Also, have you thought about support languages as URLs? It would increase shareability :)


developit:
Heh - this is good, and the language is supposed to be supported via `?lang=es` in the URL, but I broke it a couple weeks ago.  Fixing it now!

montogeek:
Oh, didn't know that

Issue #58, opened 2016-11-23
SVGs are not cached on https://preactjs.com/


abhishekisnot:
I assume https://preactjs.com/ is PWA which should work even in offline mode. But the SVG images served on preactjs.com are marked explictly with no-cache pragma. This causes the SVG images to not-render on page reload in offline mode (note: all other parts work correctly).

![screen shot 2016-11-22 at 10 57 53 pm](https://cloud.githubusercontent.com/assets/802242/20553255/d6c9b518-b107-11e6-90b0-4eaf6a72eb1c.png)

![screen shot 2016-11-22 at 11 04 06 pm](https://cloud.githubusercontent.com/assets/802242/20553270/0859e47c-b108-11e6-9420-9654b2172c0b.png)

Is this on purpose? If not, what solution do you propose?

developit:
Ah, no that's something surge.sh adds. I think the URLs need to be whitelisted for caching via OfflinePlugin /cc @nekr

NekR:
Hey @developit! How exactly you move `assets/` from `src/` to `dist/`? I doesn't seem to be done with `file-loader`. i.e. `offline-plugin` doesn't see them in webpack's output assets.

NekR:
@developit oh, I see you just copy it in here: https://github.com/developit/preact-www/blob/master/package.json#L10
If you do it outside of webpack then you will need to add each file manually to the list of `offline-plugin` caches. The easier way though it to just use `webpack-copy-plugin` which will do copy for you and populate webpack's output assets so `offline-plugin` could use it.

developit:
Ahh amazing I will definitely be switching to that.

developit:
@NekR would switching to `webpack-copy-plugin` mean I could place the external images into `additional` or `optional` sections via a `*` wildcard?

NekR:
@developit  Yeah, absolutely! 

Btw, what's this option? https://github.com/developit/preact-www/blob/master/webpack.config.babel.js#L156

There is no such option. There is `responseType: 'network-first'` in v4 though :-)

NekR:
I can make a PR to migrate to v4 btw :-)

developit:
Oh, I guess that's why it was serving from cache first! Would love a V4 PR btw.

PR #62, opened 2016-12-03
Mirgate to offline-plugin 4.5.2

NekR:
Something like this :-) I removed unnecessary options and added new ones introduced in `v4` of `offline-plugin`. For example, I removed `ServiceWorker.navigateFallbackURL` which was a temporary solution for app-shell in `offline-plugin` (it was kind of emulation AppCache's behavior with FALLBACK) and added `cacheMaps` which is official solution for app-shell in `offline-plugin` now.

@developit Please test it somehow. I did my testing and it seems fine, but I could easily miss something. You know this project better anyway :-)


developit:
Trying to test this out locally. Might deploy to a staging env with HTTPS to be sure.

NekR:
Yeah, would be good. I think I can try on `now.sh`. The we need to test is SW upgrade. It should work fine, but just-in-case testing won't hurt.

developit:
I have preact-www.surge.sh that I use as an alternate, should work there.

developit:
@NekR do you think this will solve #58?

NekR:
@developit #58 can be fixed by using `copy-webpack-plugin` with any version of `offline-plugin`. What v4 brings (in case of `preact-www`) is support of `network-first` response strategy. 

Also it's just better (4.5.4 is the latest btw).

developit:
Also the externals cover non-webpack assets as well right?

NekR:
@developit yes, `externals` is added specifically for such use cases. If you specify URLs files there then they will be caches alongside with webpack generated assets.

`externals` accepts URLs btw, not FS paths. But something like `/images/foo.jpg` or `images/foo.jpg` also works (transformed to URL, not FS checks of course).

PR #74, opened 2016-12-22
Add note about differences in context inheritance

nhunzaker:
Follow up from https://github.com/developit/preact/issues/460.


developit:
Awesome, thanks for adding this!

nhunzaker:
Sweeet, 🍬

PR #76, opened 2016-12-22
Promote fixed header to GPU

thysultan:
header gets re-painted on the main thread every time you scroll, which affects scroll performance.


developit:
Looks good to me! I also wanted to remove the opacity on `#app`, which is causing that layer (the whole page) to be composited, making the scroll pretty janky.

PR #77, opened 2016-12-30
Added Connectivity Index to the examples page.

malchata:
Connectivity Index is a Preact-powered site, and as such I feel it may be a good addition to the examples page. :) Also uses preact-render-to-string to generate server side markup.


malchata:
WORD. Thank you!

PR #82, opened 2017-01-16
add demo: preact-starter

lukeed:
reference: https://github.com/developit/preact/pull/510#issuecomment-272726562


joaolucasl:
Thanks @lukeed!

PR #86, opened 2017-01-18
Fix edit link in the footer after navigation to another page

sapegin:
Steps to reproduce:

1. Open https://preactjs.com/
2. Navigate to any docs page, say **Getting Started**
3. Click **Edit this Page** link in the footer
4. You’re editing the home page (actually the first page after site was first loaded)

### DIFF
--- a/src/components/footer/index.js
+++ b/src/components/footer/index.js
@@ -3,14 +3,14 @@ import { connect } from '../../lib/store';
 import config from '../../config';
 import style from './style';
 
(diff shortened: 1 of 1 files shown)


developit:
Nice catch - I hadn't noticed this!  Were you able to verify that it works running locally?

sapegin:
Did the same steps locally — works fine ;-)

developit:
perfect! Happy to merge 💃

PR #88, opened 2017-01-24
added missing logo and fixed min height

ritz078:


src/style/index.less
```diff
@@ -79,6 +79,7 @@ pre.highlight {
 
 content-region {
 	display: block;
+	min-height: calc(~"100vh - 256px"); // hack for less compiler which actually subtracts px from vh
```
developit: this actually relies on CSS calc, which has somewhat poor browser support.  I wonder - would #56 fix this more simply?  Cheers!

ritz078:
removed.

developit:
Cool, I'll keep looking into that footer thing.  Might actually have to end up with the calc solution if it's too annoying to use `height` haha.

PR #95, opened 2017-02-17
Add link to Instant Domain Search

hartshorne:
Our renders are much faster now, thank you!


developit:
Awesome - super fast indeed!

This page is probably better suited to Open Source projects though - would you be able to instead add your link (& logo!) to the [we-are-using](https://github.com/hartshorne/preact-www/blob/fc2c4883679f7e39860225a8d616635abb707530/content/about/we-are-using.md) page?  We're going to style that one nicely too, will be turning into a sortof "who's who" of companies using Preact 👍

PR #102, opened 2017-02-25
fixes compatability for windows

Perni1984:
When I tried to run dev mode on windows not all module could be loaded. I took some inspiration from the preact-starter-kit to make preact-www windows compatible


developit:
Does everything work on Windows with this patch?

Perni1984:
I am not sure - should the hot reload module update the content when I just change a *.md file? Because I still have to reload the page to see the updated contents.

developit:
Nope, that's not wired up - the content `.md` files are accessed via `fetch`, so they aren't tracked by webpack.

PR #105, opened 2017-03-03
Remove dead links in Big Performance section

vandrijevik:
The second paragraph in the **Big Performance** section on https://preactjs.com/ currently reads:

> It even includes extra performance features like customizable update batching, optional async rendering, DOM recycling and optimized event handling via Linked State.

The features “customizable update batching” and “optional async rendering” contain broken links that lead to 404 pages. This PR removes them, so the paragraph is simpler and contains a link to the valid guide for Linked State.

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -71,7 +71,7 @@ toc: false
     </p>
     
     <p>
-        It even includes extra performance features like <a href="/guide/configuration#debounceRendering">customizable update batching</a>, optional <a href="/guide/configuration#syncComponentUpdates">async rendering</a>, DOM recycling and optimized event handling via [Linked State](/guide/linked-state).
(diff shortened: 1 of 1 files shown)


developit:
@vandrijevik Thanks for submitting this - maybe we should keep the text, but not have it linked to anything?

vandrijevik:
@developit That sounds fine to me! I wasn’t sure about the state of those claims, but if you think they are worth mentioning I will certainly put them back.

developit:
Ah yeah - both definitely supported, just we haven't yet written documentation about them that could be linked there.  Still worth calling them out :)

vandrijevik:
@developit Sounds good to me! I added 21764f8 so that diff should look better now 😄 Are two separate commits okay, or would you prefer that I squash them into one?

developit:
Great, thanks!

PR #112, opened 2017-03-30
change some words selections in zh

huangbow:
Eliminate potential discriminative words , as "他" === he/him, "它" === it. "他将虚拟DOM与DOM本身区别开" === "He can tell the difference between Virtual DOM and DOM itself" and it's needed to correct.

### DIFF
--- a/content/lang/zh/index.md
+++ b/content/lang/zh/index.md
@@ -39,7 +39,7 @@ toc: false
     </p>
 
     <p> 
-		Preact也是web平台的一等公民，他将虚拟DOM与DOM本身区别开，注册真实的事件处理函数，很好地与其它库一起工作。
+		Preact也是web平台的一等公民，它将虚拟DOM与DOM进行diff，注册非虚拟的事件处理函数，并能很好地与其它库一起工作。
     </p>
 </section>
 
@@ -50,7 +50,7 @@ toc: false
     <h2>小体积</h2>
     
     <p>
-		大多数UI框架是相当大的，是应用程序javascript体积的大部份。Preact不同：他足够小，<em>你的代码</em>是你的应用程序最大的部分。
+		大多数UI框架是相当大的，是应用程序javascript体积的大部份。Preact不同：它足够小，<em>你的代码</em>是你的应用程序最大的部分。
     </p>
     
     <p> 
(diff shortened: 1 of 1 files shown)


developit:
Sounds like a good change @huangbow, hopefully @lcxfs1991 can review! ❤️

content/lang/zh/index.md
```diff
@@ -39,7 +39,7 @@ toc: false
     </p>
 
     <p> 
-		Preact也是web平台的一等公民，他将虚拟DOM与DOM本身区别开，注册真实的事件处理函数，很好地与其它库一起工作。
```
lcxfs1991: 原翻译没问题

lcxfs1991: 只有虚拟dom进行diff吧，有diff再更新dom

huangbow:
Thanks for the review. The original content is "It diffs Virtual DOM against the DOM itself". I would say the existing one is so abscure.

lcxfs1991:
It's my understanding that react-like framework would diff virtual dom first. It won't diff virtual dom and its corresponding real dom.  So I guess there are two options may fit author's meaning.

1. preact diffs virtual dom instead of it's corresponding dom.
2. preact diffs virtual dom for it's corresponding dom.

what's your opinion? @developit

developit:
@lcxfs1991 I'm reading this second option as "preact diffs virtual dom against it's corresponding dom" - is that correct? If so, that seems more accurate than the first.

lcxfs1991:
Gotcha. Let me fix the translation and merge the request.

huangbow:
The Diff algorithm is the true treasure of Preact. Thanks guys for helping me make it more clear.

huangbow:
I'm hesitate to take the merged translates. @developit Can you be more explicit about "preact diffs virtual dom against it's corresponding dom"? I'm afraid there's a gap between new translate and your idea. So what does Preact really diff:
1) vdom vs. vdom
2) vdom vs. corresponding dom

huangbow:
@lcxfs1991 I would be happy to read [diff code](https://github.com/developit/preact/blob/master/src/vdom/diff.js) start from line 34.

developit:
@huangbow I appreciate the concern :)
Preact definitely falls into option #2 in your question - it diffs DOM <-> VDOM.

huangbow:
@developit My original proposal (PR) was reflected what option 2. But unfortunately, @lcxfs1991 "corrected" it as vdom vs. vdom, which made me feel anxious. I hope @lcxfs1991 can stick to what I originally pulled. Thanks guys.

hassanbazzi:
Any updates on this?

PR #123, opened 2017-05-03
remove componentDidUnmount() lifecycle method

vitormalencar:
removing the the non-standard lifecycle method  `componentDidUnmount()` from the docs.


vitormalencar:
Always happy to help 😄

Issue #131, opened 2017-06-06
API documentation


wildlyinaccurate:
I often find myself going to the React docs when I forget an API. Would you be opposed to a PR that adds an "API" page with a list of function signatures and short description of each?

oren:
This would be fantastic

developit:
+1 for this, even I would find that useful.

Maybe we can start with this?

[![](https://pbs.twimg.com/media/DAhBsSPXgAMTJh2.jpg)](https://twitter.com/_developit/status/867015201383731200/photo/1)

wildlyinaccurate:
Great, I'll try and make a start later this week! I'll definitely start with that, and I also wanted to document `render()` because I _always_ trip up on what to put in the second and third args to make it replace a server-rendered node.

oren:
yes! the render function can drive me nuts. I'll buy you a beer if we ever meet.

PR #132, opened 2017-06-06
Get started on API Reference documentation

wildlyinaccurate:
Following the discussion on #131, I've started on an _API Reference_ page. I'm opening the PR now to promote early feedback ⭐️ 

I realise that this page will probably duplicate some of the content on other pages, but I'm not sure that's necessarily a bad thing. Hopefully I'm not the only person who heads straight for the API Reference page whenever I need to remember how to do something!

My todo list for this page so far is:

- [x] Document the return types for each `Component` lifecycle method.
- [x] An example of a stateless functional component.
- [x] Some examples for `render()` including how to replace server-rendered content.


developit:
Looks awesome so far!  Should we wait to merge, or get this out in its current state?

wildlyinaccurate:
I've had a busy couple of weeks. I'd intended to write a bit more before we merge it. I also wanted to ask you what position in the navigation this should take. Maybe I should move it to the bottom?

developit:
I was actually thinking it could entirely replace the "Lifecycle Methods" link, especially if we copied that table into this page.

wildlyinaccurate:
Forgot to ping you last week, @developit. What do you think of this? Anything you'd like added/removed/tidied up?

wildlyinaccurate:
I've done my best to transfer the lifecycle docs for Zhongwen and Portuguese but it does leave the page a little empty compared to the English version.

developit:
@wildlyinaccurate I changed two things (split into 2 commits) - let me know if that makes sense.  Otherwise I think we're good to go!

wildlyinaccurate:
Thanks for clarifying those sections! Makes much more sense now. I'm happy if you are 🎉

developit:
Yup!

PR #135, opened 2017-06-30
Added missing link for React lifecycle methods

PandaWhisperer:
Found a missing link in the [Getting Started](https://preactjs.com/guide/getting-started#components) document. 

Possibly this was meant to link [here](https://facebook.github.io/react/docs/state-and-lifecycle.html)?


developit:
I think it was meant to link [here](https://preactjs.com/guide/lifecycle-methods), but we're changing that documentation in #132.

PandaWhisperer:
@developit does that mean this PR is no longer necessary? If so, please go ahead and close it.

PR #136, opened 2017-07-05
Add Node Monkey Patching react to preact

azizhk:
This is something that we have done on our server implementation. Just creating a PR as well as opening a discussion as to if there are any better alternatives.

- We can use [yarn's alias](https://twitter.com/sebmck/status/873958247304232961/photo/1) but that does not work as your dependencies in node_modules will still have reference to react and not preact. 
- We can use a babel plugin like https://github.com/tleunen/babel-plugin-module-resolver to replace preact to react but I could not get it to work. I think babel ignores transpiling node_modules or I have to set something to get it to transpile node_modules.

Open to grammar, capitalization changes if this gets accepted.


developit:
Looks good @azizhk.  Do you think it would be worth mentioning [module-alias](https://www.npmjs.com/package/module-alias) here too?  I believe it works similarly to the example, albeit harder to see what's happening to newcomers.

azizhk:
Yup, module-alias looks like a better solution and we might not have to require react at all.
Give me some time, I'll update the PR before EOD.

PR #146, opened 2017-07-17
Improve page to match original style

davi-mbatista:
Matching the old pt-br design with the current.


developit:
I'll make sure to keep this updated for the other translations too :)

PR #156, opened 2017-08-19
Add Selly to we are using

Sexual:
Selly is a heavy advocate of Preact


developit:
Awesome! Glad to hear you're finding it useful :D

PR #169, opened 2017-09-07
Update German translation

ShalokShalom:
The current German translation looks like translated by Google, which is especially since your original English source use such phrases like "It diffs Virtual DOM against the DOM itself" and "Component paradigm" - those ones are insanely complicated to translate.  

I see the option to use a simple description in the Original, instead of the fancy tech-speaking. 
This counts probably for all the other languages as well.

This commit here patch some of the more obvious and easy to fix things.


ShalokShalom:
Maybe we want to add a link to explain what a [Virtual DOM](http://reactkungfu.com/2015/10/the-difference-between-virtual-dom-and-dom/) is actually.

developit:
@ShalokShalom I was also curious if it would make sense to change the homepage title from:
"Eine Bibliotek der anderen Art."
to:
"Eine lib der anderen art."

Would that be more colloquial?

jeyj0:
@developit That would not be more colloquial. It sounds more Google-translated again. 😉  Instead use "Bibliothek" - which does not contain a spelling mistake (the "h" is missing on the website). 😄

ShalokShalom:
I agree with jeyj0

PR #173, opened 2017-09-19
Added new Demo/Example

tomasswood:
Added in a new Demo/Example for a Preact Homepage Generator


tomasswood:
Awesome, thanks! :D

PR #179, opened 2017-10-22
Port threeJS codepen over into component.

gl2748:


reznord:
This is a good demo and probably can be ported to a template for preact-cli.

Then we can add it in the preact-cli's wiki

PR #185, opened 2017-11-30
Add Babel 7 version in getting-started

xtuc:
Saw that while reviewing the french version. We didn't change the API in Babel 7.


developit:
Good call.

PR #189, opened 2017-12-18
Added The DEV Community logo to we-are-using page

benhalpern:
Long-time project follower/fan, happy to say we now have Preact in production. Looking forward to helping spread the good word of Preact more and more.


developit:
this makes me so happy :D

PR #205, opened 2018-02-09
Update differences-to-react.md: Update hash

Dorumin:
The fragment identifier for the section was changed from react.children to just reactchildren (or maybe it was never react.children, I dunno)


developit:
Thanks!

PR #206, opened 2018-02-18
Remove devpad.io (dead link)

targumon:


targumon:
BTW, @developit, I was reviewing this page because I'm about to develop a web widget using Preact and was asked by my coworkers if this scenario is 'battle tested'. Unfortunately I can't find any company who already did that 😕.

developit:
@targumon oh, there are tons! Qantas, Algolia, FT and NYT come to mind - they all use Preact for that use-case.

PR #221, opened 2018-05-22
Add bakkenbaeck's svg logo

Jinex2012:


reznord:
Would be great if you can format the SVG, instead of putting in a single line

PR #226, opened 2018-07-12
Update context api documentation

tejohnso:
Resolves [#1167](https://github.com/developit/preact/issues/1167), clarifying support for old React Context API rather than the new API.


reznord:
@marvinhagemeister IIRC, for the new context API, we can use [preact-context](https://github.com/valotas/preact-context) right?

PR #240, opened 2018-08-29
getting-started.md - Broken link

alex-page:
In my recent PR #238 I was not sure how the link was working as it was just `[preact-boilerplate]` without the markdown link brackets. This seems to have broken the link, so I have added the url with the conventional markdown syntax.


developit:
Thanks for catching that!

PR #244, opened 2018-09-12
docs: Add CDN link to mjs bundle.

pl12133:
Related to https://github.com/developit/preact/issues/1207.


developit:
Thanks Peter! I'm going to add unpkg links as well.

PR #245, opened 2018-09-14
Completed api-reference.md translation to pt-br

guaiamum:


reznord:
Trusting @caina-jumia

PR #249, opened 2018-10-13
conditional rendering example

devalexandre:
make a conditional rendering example


developit:
I think this content would make more sense in a larger "getting started with Virtual DOM" guide, rather than being a dedicated page. Let's wait until that guide is underway and merge this into it, since merging now would break URLs in the near future.

hassanbazzi:
@developit Any latest thoughts on this?

developit:
@hassanbazzi I don't think we want a full top-level page for this, but I would love to see this content make it's way into a proper "starting from scratch" guide.

PR #250, opened 2018-10-19
[Do Not Merge] - Move/rename 'Differences to React' to 'Differences from React'

amatecha:
Minor change to rename this page.  Not essential by any means!  As to the merits of making this change, trying other forms of the base word "differ" makes the awkwardness of "to" readily apparent -- you don't say "this differs to this other thing", you say "this differs from this other thing", for example.  :)


developit:
I agree with the name change, but this will break URLs in a way we don't have a good fix for yet.

Perhaps we can slate it in for when we refactor the information architecture?

Note to maintainers - do not merge this (yet)!

rschristian:
We can land this without breaking URLs now (or at least, we can redirect without too much issue), though I'm not sure I see a great need after so many years.

As such, will close this on out.

PR #270, opened 2019-02-13
Guard the usage of localStorage to catch exceptions.

molily:
Hello! I’ve configured my browser to block cookies unless allowed per domain. Access to localStorage throws an exception in this case. This blows up many web sites that render content using JS and do not guard the use of localStorage with a `try {} catch () {}`. Including preactjs.com:

screenshot 2019-02-13 at 09 50 40

This change introduces helper functions for accessing localStorage wrapped in try/catch.

Let me know if you find this useful.


src/store.js
```diff
@@ -22,13 +23,13 @@ export default () => {
...
 	for (let i=SAVE.length; i--; ) saved[SAVE[i]] = state[SAVE[i]];
-	localStorage.state = JSON.stringify(saved);
+	localStorageSet('state', JSON.stringify(saved));
 }
 
 function getSavedState() {
 	let state;
 	try {
-		state = JSON.parse(localStorage.state);
+		state = JSON.parse(localStorageGet('state'));
```
molily: This line is already wrapped in a try-catch, so it was safe already, but i’ve used localStorageGet here for consistency.

PR #272, opened 2019-02-19
add CLI readme

hassanbazzi:


hassanbazzi:
@developit merge this one?

reznord:
@hassanbazzi Looks good to me for a basic info, but wondering if we need to add more info about the CLI including what the commands do.

/cc @developit @lukeed

PR #275, opened 2019-02-21
Add Mopinion to companies using Preact

Anwardo:
We rebuilt our survey tool with Preact and achieved great speed improvements and reduced filesize.


hassanbazzi:
Thank you! Check out Preact X as well :)

PR #276, opened 2019-02-22
willComponentUpdate missing from example on lifecycle methods

craigdanj:
I noticed this lifecycle method was missing from the documentation. Dunno if we wanna add it in?


hassanbazzi:
It's actually componentWillUpdate. It's in the reference. :)

craigdanj:
Well this is embarrassing... musta been tired when I did this.

PR #281, opened 2019-04-01
Fix webpack.js.org alias link

developit:
see https://twitter.com/johnmosesman/status/1112527189038301185

### DIFF
--- a/content/guide/switching-to-preact.md
+++ b/content/guide/switching-to-preact.md
@@ -35,7 +35,7 @@ to redirect any imports/requires looking for `react` or `react-dom` with `preact
 
 #### Aliasing via Webpack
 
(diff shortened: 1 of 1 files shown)


johnmosesman:
Hey thanks!

It might also be worth adding a new section for webpack 4. The docs currently show:

```
"alias": {
  "react": "preact-compat",
  "react-dom": "preact-compat"
}
```

But what I actually needed was to alias the module not the file name:

```
alias: {
  React: "preact-compat",
  ReactDOM: "preact-compat"
}
```

developit:
hmm - @johnmosesman is it possible you're using a `umd` output target? Normally module filenames are lowercase.

PR #283, opened 2019-04-05
(docs) - state hooks basics

JoviDeCroock:
I've made a first draft to see if this style conforms to how we see the docs.


JoviDeCroock:
This is the first draft, in slack we agreed on making the hooks chapter by subchapter (PR by sub). Mainly want to know if this comes across to informal and if we want to see some things differently.

I also will see if I can find something for interactive code samples, now they're static examples

PR #285, opened 2019-04-14
Fix bug in example code

simoncoggins:


mehmetkose:
seems not a bug

simoncoggins:
Ah, sorry my mistake, I was missing the arguments for the render() method when I copied the example.

PR #292, opened 2019-07-08
Minor Design UX improvements

marvinhagemeister:
This PR makes minor changes to the css to make our site easier to read.

- Increase font size from `15px` to `16px`. It's not much but makes the font easier to read on my screen.
- Increase `h2` size to make it easier to scan for the eyes
- Change text link styles to be more upbeat and visually more distinct from their surroundings
- Make blockquotes more visually appealing by using colors. This better differentiates them from the body text and hints to the user in a visual way that it may contain some useful information

The changes are best illustrated via a screenshot.

Before:

preact-typo-before

After:

preact-typo-after

### DIFF
--- a/src/style/markdown.less
+++ b/src/style/markdown.less
@@ -1,9 +1,14 @@
 .markup, .markdown {
-	& > *:first-child {
-		margin-top: 0;		// !important
+	& > * ~ * {
+		margin-top: 1.25rem;
 	}
-	& > *:last-child {
-		margin-bottom: 0;	// !important
+
+	p a {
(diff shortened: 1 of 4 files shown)


src/style/markdown.less
```diff
@@ -1,9 +1,16 @@
 .markup, .markdown {
-	& > *:first-child {
-		margin-top: 0;		// !important
+	& > * ~ * {
```
marvinhagemeister: Also known as the [lobotomized owl selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/)

PR #295, opened 2019-07-08
Fix donation corner badge text being cut off

marvinhagemeister:


marvinhagemeister:
Closing it looks different than in my local dev setup :S

PR #296, opened 2019-07-08
Bugfix: remove custom ServiceWorker so workbox functions correctly

developit:


marvinhagemeister:
Looks like we need to run `npm install` again to update the `package-lock.json`

PR #298, opened 2019-07-08
Check if browser supports SW before using them

marvinhagemeister:
Not all browsers support Service-Workers, so this PR checks for them before trying to use them. I've run into this in my local dev setup which doesn't use https.


marvinhagemeister:
That's true, but without it, it will automatically pause the debugger due to the uncaught exception :see_no_evil:

PR #300, opened 2019-07-09
Sharing Footer

developit:
Automatically rotate through top contributors in footer instead of always putting my name.

Screen Shot 2019-07-09 at 9 28 35 AM

### DIFF
--- a/src/components/footer/index.js
+++ b/src/components/footer/index.js
@@ -7,7 +7,7 @@ import style from './style';
  *  To update this list, on https://github.com/preactjs/preact/graphs/contributors run:
  *	$$('.contrib-person [data-hovercard-type="user"]:nth-of-type(2)').map(p=>p.textContent).filter(x => !/-bot$/.test(x)).join(' ')
  */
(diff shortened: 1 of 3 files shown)


src/components/footer/index.js
```diff
@@ -3,16 +3,34 @@ import { connect } from 'unistore/preact';
...
 	setLang = e => {
 		this.props.store.setState({ lang: e.target.value });
 	};
 
-	render({ lang, url = location.pathname }) {
+	render({ lang, url = location.pathname }, { contrib }) {
 		let path = url.replace(/\/$/,'') || '/index';
 		if (lang) path = `/lang/${lang}${path}`;
-		let editUrl = `https://github.com/developit/preact-www/tree/master/content${path}.md`;
+		let editUrl = `https://github.com/preactjs/preact-www/tree/master/content${path}.md`;
```
marvinhagemeister: Good catch 👍

developit: heh - I had noticed the star count API call had started redirecting, saves us the overhead.

PR #302, opened 2019-07-09
Make table of contents always visible

marvinhagemeister:
This PR completely rewrites the table of contents area. It's now always visible on the left and will highlight the current heading. On mobile it's moved behind a toggle on the lower right corner of the screen. On top of that I slightly modified the heading styles and made all link colors use our brand color.

preact-toc-desktop

preact-toc-mobile

### DIFF
--- a/src/components/controllers/page/index.js
+++ b/src/components/controllers/page/index.js
@@ -4,6 +4,7 @@ import cx from '../../../lib/cx';
 import ContentRegion from '../../content-region';
 import config from '../../../config';
 import style from './style';
+import Toc from './table-of-content';
 
 const EMPTY = {};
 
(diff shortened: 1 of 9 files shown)


marvinhagemeister:
There are cases where the table of contents is not refreshed properly, but that's something I'd rather tackle in a separate PR. This PR is just for styling mostly.

developit:
Looks awesome. Would you be open to making the mobile FAB button red with white text?

ForsakenHarmony:
🤔 what's going on there

marvinhagemeister:
@ForsakenHarmony should be fixed now, can you check again?

@developit sure thing :+1:

PR #304, opened 2019-07-11
Add prettier

marvinhagemeister:
This PR adds prettier and enables it for `.js`, `.json` and `.less` files inside the `/src` directory.


package.json
```diff
@@ -20,7 +21,8 @@
     "extends": "developit",
```
JoviDeCroock: Wouldn't it be better to remove the already present eslint config when we are using prettier?

marvinhagemeister: The eslint preset contains many linting rules not related to formatting. That's why I left it there 🙂

JoviDeCroock: All right, disregard my comment then :)

PR #305, opened 2019-07-11
Simplify Hero/Jumbotron

marvinhagemeister:
This PR simplifies our header a bit to make it look "cleaner". I'll expect it to undergo more changes in the coming week, but this is a good starting point we can work on :+1: 

preact-hero

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -8,17 +8,10 @@ import Footer from './footer';
 /*global ga*/
 
 let store = createStore({
-  location: location.pathname
+	url: location.pathname
 });
 
 export default class App extends Component {
-	componentDidMount() {
-		if (typeof window !== undefined) {
-			// This triggers a second render :/
-			store.setState({ url: window.location.pathname });
-		}
-	}
-
 	handleUrlChange({ url }) {
 		let prev = store.getState().url || '/';
 		if (url !== prev && typeof ga === 'function') {
@@ -27,7 +20,8 @@ export default class App extends Component {
 		}
 	}
 
-	render({ url }) {
+	render() {
+		let url = store.getState().url;
 		return (
 			<Provider store={store}>
 				<div id="app">
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -7,7 +7,9 @@ import Footer from './footer';
 
 /*global ga*/
 
-let store = createStore();
+let store = createStore({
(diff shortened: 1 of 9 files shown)


src/components/app.js
```diff
@@ -10,6 +10,13 @@ import Footer from './footer';
 let store = createStore();
 
 export default class App extends Component {
+	componentDidMount() {
+		if (typeof window !== undefined) {
+			// This triggers a second render :/
+			store.setState({ url: window.location.pathname });
```
marvinhagemeister: Need to find a way to initialize the store with the correct location instead of updating it later.

src/components/app.js
```diff
@@ -10,6 +10,13 @@ import Footer from './footer';
 let store = createStore();
```
developit: `location` is available during prerendering ([source](https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/prerender.js#L14)):

```suggestion
let store = createStore({
  location: location.pathname
});
```

src/components/app.js
developit: can do it when initializing the store: https://github.com/preactjs/preact-www/pull/305/files#r302571204

developit:
Love it. This reminds me of a design proposal @zouhir made a while back.

src/components/app.js
marvinhagemeister: Sweeet!! 💯

marvinhagemeister:
@zouhir Any chance you still got the design somewhere? Maybe we can reuse some pieces :slightly_smiling_face:

src/components/app.js
```diff
@@ -18,11 +20,12 @@ export default class App extends Component {
 		}
 	}
 
-	render({ url }) {
+	render() {
+		let url = store.getState().url;
```
JoviDeCroock: Personally I like to get this in the HOC instead of drilling it down (nitty) but this maybe isn't the time and place to discuss this.

marvinhagemeister: Yeah I agree. It should be passed through `context` instead. 8.x doesn't have the createContext API and I think this will be a lot easier to do once we switch to X with hooks here 👍 I'm kinda postponing cleanups like that until the switch.

JoviDeCroock: Agreed

PR #309, opened 2019-07-13
Prettier + Linting + jest fixes

marvinhagemeister:
Due to an incorrect glob path we missed a few files when switching to prettier. On top of that I added a `prettierignore` file because otherwise vscode would format everything. This may have been true for other editors too.


marvinhagemeister:
hm the puppeteer tests seem to be a bit flaky :/

marvinhagemeister:
Ok, found the issue. It seems that `worker-loader` is super strict about code formatting :man_shrugging:

marvinhagemeister:
Note it may be easier to review this PR by commit. The changes caused by prettier formatting is always a bit noisy :slightly_smiling_face:

marvinhagemeister:
Merging, because all my other PRs are based on it and it mainly cleans up build tools :+1:

PR #313, opened 2019-07-14
Remove unused markup feature

marvinhagemeister:


developit:
Seems fine - I don't think this will impact size much. I'd originally left it in because I had planned to move the REPL over to being a page with frontmatter, but I think we can do that just as easily using markdown (since we could just use HTML in the markdown file anyway).

PR #327, opened 2019-07-22
Switch to CSS Custom Properties for theming

marvinhagemeister:


developit:
Test isn't flakey, it's tripping because >40% of the CSS on the page is unused. I can try to bring that down by lazy-loading the theme CSS.

marvinhagemeister:
@developit I'm not sure if it's just the theme or if these savings can't be done elsewhere. I'm curious to see what the biggest contributors to size are in the CSS.

marvinhagemeister:
Retried the CI and it's green now :man_shrugging:

PR #329, opened 2019-07-23
Progressive hydration

developit:
cheap way to get a netlify preview.


marvinhagemeister:
This looks great and I'm excited about an even faster user experience :tada: On the instance netlify deployed I'm not able to open the menus anymore (same for mobile). No error is thrown in the console but something seems to be going on.

developit:
Hmm, weird - I'm able to open them on the deploy preview. Seeing some funky stuff with the ToC being hidden though.

marvinhagemeister:
Just checked again and everything seems to work now. Not sure what was up with my phone. Clicking on menus works fine for me again (probably something on my end).

marvinhagemeister:
Brought the latest changes from master into this PR to get the CI fix and the correct lockfile.

developit:
its READY!

marvinhagemeister:
Super awesome! Just clicked through a production build and it's as snappy as it has ever been! :+1: 

The only issue I found so far is that the `Hydrators` render an empty tag when in development mode (sidebar missing etc). I was able to track it down to the toc never being set because that happens only during prerendering.

developit:
@marvinhagemeister ToC works in dev mode now.

PR #332, opened 2019-07-24
Add different doc versions

marvinhagemeister:
This PR adds a way to switch between different doc versions. The version is always part of the url, like `/guide/v10/getting-started`. It can be changed via the dropdown menu inside the sidebar. The UI was previously already present, but didn't actually do anything.

**Move docs menu into the sidebar**

When I played around a bit with the version switching I noticed that you always have to reorient yourself. Before moving the docs menu into the sidebar we would expect the user to click the "Guide" button in the header. This seems wrong to me and I'd rather have the menu somewhere close to the dropdown. One benefit of moving the menu there is that now the user can visually see that something changed. It's much harder to convey when most of the changes are hidden behind a header menu.

**Tasks:**
- [x] Make dropdown do something
- [x] Add versioned docs
- [x] Move docs menu into sidebar
- [x] Code-Cleanups

preact-sidebar-nav


marvinhagemeister:
Shoot I wanted to open it just as a draft, but clicked to quickly :man_facepalming: I can't find an option to mark it as a draft again.

developit:
I really like the menu in the sidebar, babeljs.io-style. Do you think we should still try to do something for a table of contents?

marvinhagemeister:
@developit Yeah we should have some form of a TOC. Tried a few approaches:

**1. TOC as a third column**

Both babel and jest (and probably many more) place the TOC in a third column right next to the content. I'm guessing that this is the default for docusaurus on which both sites are built on. The obvious question here is: Is the TOC important enough to
warrant its own column? It leads to a lot of wasted space and makes responsive styling a bit more difficult for not much gain. In my opinion it's too reminiscent of typical layouts that were used for pretty much any page (at least here in Germany) around 2003-2009. Just feels dated to me.

**2. TOC under menu entry in sidebar**

This one is great in theory but imo can be somewhat confusing once you introduce sub-page instead of the flat list of documents we have right now. It's hard to distinguish from what's a real menu entry and what's the TOC. I don't it just feels like the TOC hasn't enough of a standing to be included there.

**3. Put TOC into content**

Instead of putting it in a special place, why not make it part of the content itself? With that we can write a generic introduction paragraph for the page and list the TOC followed by the content afterwards. I like this one a lot as it just feels like it's part of something and not just another thing to skim over. The cool thing about this one is that many markdown plugins for editors or IDE autogenerate it. So we can skip generating it on the client for a tiny perf win.

localhost_8080_guide_v10_differences-to-react (1)

PR #334, opened 2019-07-24
Fix CI flakyness

marvinhagemeister:
This PR fixes all the weird test failures on my machine. There were several issues with our setup:

- One test ran before all scripts are loaded
- Code coverage always included external resources like google analytics
- Code coverage sometimes contains duplicate coverage information for the same resource (not sure why)
- We didn't wait for the server to boot up completely before starting our test suite (one **must** set the `port` option for `jest-puppeteer`. Only found that buried in a GitHub issue)
- Upgrade `puppeteer` + `jest-puppeteer` (just to be sure)
- We didn't caught errors when any of the requests to the GitHub API failed

### DIFF
--- a/test/browser/homepage.test.js
+++ b/test/browser/homepage.test.js
@@ -2,6 +2,17 @@ jest.setTimeout(10000);
 
 const URL = 'http://localhost:4444';
 
+/**
+ * Filters out external resources and duplicate coverage information
+ */
+function getUniqueResources(arr) {
+	const seen = new Map();
+	arr
+		.filter(x => x.url.startsWith('http://localhost'))
+		.forEach(x => seen.set(x.url, x));
+	return Array.from(seen).map(x => x[1]);
+}
+
 async function getCoverageData(customPage = page) {
 	const [js, css] = await Promise.all([
 		customPage.coverage.stopJSCoverage(),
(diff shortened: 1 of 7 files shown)


test/browser/homepage.test.js
```diff
@@ -2,6 +2,17 @@ jest.setTimeout(10000);
 
 const URL = 'http://localhost:4444';
 
+/**
+ * Filters out external resources and duplicate coverage information
+ */
+function getUniqueResources(arr) {
```
developit: wow, good to know. I'm guessing this is caused by preloads? yikes.

test/jest-puppeteer.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
 		devtools: process.env.HEADLESS === 'false'
 	},
 	server: {
-		command: 'PORT=4444 npm run server'
+		command: 'PORT=4444 npm run server',
+		port: 4444
```
developit: heh - I had this locally and assumed it wasn't doing anything. My bad!

marvinhagemeister: No worries, I just which the `jest-puppeteer` docs would mention that this is required :man_shrugging:

test/browser/homepage.test.js
marvinhagemeister: Yeah not sure why they are accessed twice. It's something we should keep an eye on.

PR #335, opened 2019-07-24
Only trigger CI once per push

marvinhagemeister:


ForsakenHarmony:
I guess it won't run if you don't open a pr then

marvinhagemeister:
@ForsakenHarmony yeah that's true. Personally I think it's a good compromise because we can use drafts for other cases, but let's see what everybody else thinks.

PR #341, opened 2019-07-26
Remove Linked State in v10 docs

marvinhagemeister:
Thanks for all the fish!

(PR is based on #332 . I'll mark it as ready once that is merged)


marvinhagemeister:
On second thought we should probably keep it under some `Class-Component` section. It may still be useful for projects where hooks can't be used.

PR #355, opened 2019-07-29
Replace highlightjs with prism

marvinhagemeister:


yarn-error.log
```diff
@@ -0,0 +1,135 @@
+Arguments:
```
developit: delete? haha

marvinhagemeister: Whoops😅

PR #357, opened 2019-07-30
Bring back dynamic TOC generation

marvinhagemeister:
**Tasks:**

- [x] Replace static toc with component in `*.md` files
- [x] Extract TOC component from page
- [x] Display current TOC (somehow it's either the previous one or nothing)
- [x] ~~Integrate with Hydrator (would that make sense?)~~ - I don't think that's doable


src/lib/markdown.js
```diff
@@ -14,12 +16,28 @@ const OPTIONS = {
 export const markdownToHtml = md =>
 	CACHE[md] || (CACHE[md] = marked(md, OPTIONS));
 
+// Make sure that the markup component actually re-renders when the TOC is
+// updated. This obviously pretty a gross hack, so don't copy that.
+// TODO: Add support for stateful hooks in preact-markup.
```
marvinhagemeister: I'll check tomorrow if I can find a better solution for this.

PR #369, opened 2019-08-03
Remove flatmap + unused dependencies

marvinhagemeister:


developit:
Looks like this broke prerendering. TBH I never use flatMap, maybe it's slightly different than the reduce here? I think @hassanbazzi added it in order to account for the nested route definitions within `guide.routes`: https://github.com/preactjs/preact-www/blob/master/src/config.json#L42

src/prerender.js
```diff
@@ -1,11 +1,10 @@
...
 // const { resolve } = require('path');
-const flatMap = require('flatmap');
 const config = require('./config.json');
-
-const routes = flatMap(config.nav.concat(config.docs), arr =>
-	arr.path ? { path: arr.path, name: arr.name } : arr.routes
-);
+// eslint-disable-next-line
+const routes = config.nav.concat(config.docs).reduce((acc, arr) => {
+	return acc.concat(arr.path ? { path: arr.path, name: arr.name } : arr.routes);
```
developit: I would have thought this would work too, but I don't think `reduce()` invokes the reducer callback on newly-added items in the accumulator, so `arr.routes` being concatenated onto it just concats them inline. The weird thing is that, assuming they're in the right format, that... should work?

hassanbazzi:
Flatmap is magic because it works with nested children. Why kill it homie?

PR #374, opened 2019-08-04
Update Enzyme testing documentation for Preact v10

robertknight:
This PR updates the Enzyme testing documentation for Preact v10 to use the "official" Preact adapter (enzyme-adapter-preact-pure). The new guide mostly applies to Preact 8 as well, although the examples will need to be different since the v10 guide uses an example with hooks. I will look at updating the Preact v8 docs separately.

Changes in detail:

- Update the Enzyme testing documentation to recommend the
  enzyme-adapter-preact-pure package which is part of the Preact project.
  It is easier to set up and avoids several of the limitations of the
  previous adapter.
- Expand the into to explain what Enzyme is in a little more detail, how
  it supports different React versions and React-like libraries and the
  different execution environments (browser with Karma, Node with jsdom)
  it supports
- Rewrite the example to use a simpler component, with the complete definition
  of the component included in the example
- Add a section on how Enzyme works at a high level, since understanding the
  conceptual model is helpful for understanding why other aspects of it work as
  they do
- Add a section on Enzyme's different rendering modes and when to use them
- Add a section on the `act` function from `preact/test-utils` and explain when
  it is needed and how to use it
- Add a section on differences between Enzyme + React and Enzyme + Preact,
  which explains in particular the different way that "shallow" rendering and
  state updates work

### DIFF
--- a/content/en/guide/v10/unit-testing-with-enzyme.md
+++ b/content/en/guide/v10/unit-testing-with-enzyme.md
@@ -6,7 +6,20 @@ description: 'Testing Preact applications made easy with enzyme'
 
 # Unit Testing with Enzyme
 
-React provides a `react-addons-test-utils` module for testing components, and Airbnb's `enzyme` takes this concept further - incorporating multiple rendering modes and other useful features. Testing preact components using `enzyme` is possible thanks to the `preact-compat-enzyme` module, which implements the necessary internal React properties on top of `preact-compat`.
+Airbnb's [Enzyme](https://airbnb.io/enzyme/) is a library for writing
+tests for React components. It supports different versions of React and
+React-like libraries using "adapters". There is an adapter for Preact,
+maintained by the Preact team.
+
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/unit-testing-with-enzyme.md
```diff
@@ -16,78 +29,197 @@ React provides a `react-addons-test-utils` module for testing components, and Ai
...
+
+```js
+import { configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-preact-pure';
+
+configure({ adapter: new Adapter });
+```
+
+For guidance on using Enzyme with different test runners, see the
+[Guides](https://airbnb.io/enzyme/docs/guides.html) section of the Enzyme
```
marvinhagemeister: +1 That's so much better than duplication their docs :100:

robertknight:
Thanks for the super-swift reviews @marvinhagemeister and @JoviDeCroock :)

PR #376, opened 2019-08-06
Fix broken site in IE11

marvinhagemeister:
- [x] Fix CSS layout issues
- [x] Disable theming
- [x] Progressively enhance experience with `@supports`
- [x] Fix SVGs cutoff due to missing `viewBox` attribute
- [x] Fix fallback CSS Variable values not working (variables must be in import graph at compile time)
- [x] Polyfills.

This PR makes our site work well with IE11. The only thing I still need to figure out is how `preact-cli` injects polyfills. In theory the custom polyfills in this PR shouldn't be needed, but somehow IE throws exceptions all over the place without it. Need to investigate a bit more.


marvinhagemeister:
After a bit of investigation it seems like something is failing with the `Promise` polyfill. Not sure what exactly triggers it, as it's hard to get a decent stack trace out of IE11...

PR #377, opened 2019-08-06
Update preact-cli-plugin-netlify

developit:
@vinaypuppal merged my changes so we don't need a fork anymore 🎉


marvinhagemeister:
Looks like we need to update the `package-lock.json` file too :+1:

PR #379, opened 2019-08-08
Hoist content fetching, Workerization

developit:
Might be a rough rebase, sorry for taking so long on this. Right now only syntax highlighting is off-main-thread. I extracted the markdown parsing Workerization into a follow-up PR (TBD).


src/components/code-block/index.js
```diff
@@ -0,0 +1,102 @@
+import { useState, useMemo, useRef, useEffect } from 'preact/hooks';
+import { Link } from 'preact-router';
+import cx from '../../lib/cx';
+import PrismWorker from 'workerize-loader?name=prism.[hash:5]!./prism.worker';
+
+const { highlight } = PRERENDER ? require('./prism.worker') : new PrismWorker();
+
+function useFuture(initializer, params) {
```
developit: This hook abstracts away the complicated parts of using a maybe-promise API like `highlight()`. This means during SSR, hydration and cached navigation, the first render can be fully syntax-highted synchronously without having to roundtrip through the Worker and cause a second render.

src/components/controllers/page/index.js
```diff
@@ -42,56 +43,63 @@ export function useDescription(text) {
...
+	let [meta, setMeta] = useState(hydrated ? bootData.meta : undefined);
+	if (!meta) meta = (hydrated && bootData.meta) || {};
 
+	const lock = useRef();
 	useEffect(() => {
 		if (!didLoad) {
 			setLoading(true);
 		}
-	}, [getContent(route)]);
+		const contentId = getContentId(route);
```
developit: This ensures quickly switching to new routes doesn't flicker old content in. Triggering a page load invalidates any in-flight requests for content.

src/components/controllers/page/index.js
```diff
@@ -116,19 +124,26 @@ export function usePage(route, lang) {
...
 	const layout = `${meta.layout || 'default'}Layout`;
-	const name = getContent(route);
+	const name = getContentId(route);
 
-	const isReady = !loading && urlState.toc != null;
+	const isReady = !loading;
+
+	// workaround: we toc in the store in order for <table-of-contents> to pick it up.
+	if (meta.toc && ctx.store.getState().toc !== meta.toc) {
+		ctx.store.setState({
```
developit: required for ToC to be in sync on first render

src/components/table-of-contents/index.js
```diff
@@ -16,16 +16,6 @@ export default function Toc() {
 			: cache.current || [];
 	}, [toc]);
 
-	// We'll set the current height of the last toc and keep it around until
```
developit: no longer needed! :D

src/lib/content.js
```diff
@@ -0,0 +1,171 @@
...
+			ext = 'md';
+			return fetch(`${path}/${r.status}.md`);
+		})
+		.then(r => r.text())
+		.then(parseContent)
+		.then(applyEmojiToContent)
+		.then(data => {
+			data.fallback = fallback;
+			return data;
+		});
```
marvinhagemeister: Should we add a `catch` handler to clear the cache when the request failed? A hard reload would do the same thing so I'm not sure if this is necessary.

src/components/code-block/index.js
```diff
@@ -0,0 +1,102 @@
...
+	);
+	const repl =
+		(lang === 'js' || lang === 'jsx') &&
+		code.split('\n').length > 2 &&
+		props.repl !== 'false';
+	const canHighlight = !!pending || !error;
+	const html =
+		(canHighlight && highlighted) ||
+		code.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+	const htmlObj = useMemo(() => ({ __html: html }), [html]);
```
marvinhagemeister: We could theoretically skip the `useMemo` here because we compare the html property during diffing:

https://github.com/preactjs/preact/blob/25b15fb1e352f5dbb863a61b5e911d3f264b6da2/src/diff/index.js#L232

src/components/code-block/prism.worker.js
```diff
@@ -0,0 +1,9 @@
+import * as prism from '../../lib/prism';
+
+export function highlight(code, lang) {
+	if (lang == 'sh') lang = 'bash';
```
marvinhagemeister: Good catch 👍

src/components/header/search.js
```diff
@@ -34,7 +36,7 @@ export default class Search extends Component {
 				docsearchInstance = docsearch({
 					apiKey: config.docsearch.apiKey,
 					indexName: config.docsearch.indexName,
-					inputSelector: `#${this.id}`
+					inputSelector: this.input.current
```
marvinhagemeister: Note to self: check if passing a node instead of a selector works with docsearch

src/components/table-of-contents/index.js
marvinhagemeister: I'm so glad that this hack is gone 👌

src/lib/content.js
```diff
@@ -0,0 +1,171 @@
...
+	if (!pendingEmojiProcessor) {
+		pendingEmojiProcessor = import(
+			/* webpackChunkName: "emoji" */ './gh-emoji'
+		).then(({ replace }) => (processEmojis = replace || String));
+	}
+	return pendingEmojiProcessor.then(processEmojis => processEmojis(content));
+}
+
+/* generate ToC from markdown */
+function generateToc(markdown) {
```
marvinhagemeister: So good 😍

marvinhagemeister:
@developit Maybe you forgot to commit `markdown.js`?  There is an import listed in the changes but I can't find the PR in any commits here.

FYI: Merged the current `master` into this branch :tada:

src/components/header/search.js
marvinhagemeister: Just checked and it works :tada:

developit:
Ah crap, I thought it was unused.

src/components/code-block/prism.worker.js
developit: I'm actually not 100% sure if this is even needed. I couldn't find an alias for it in Prism, but we are also missing the `"shell"` lang definition there.

src/components/header/search.js
developit: lol yeah their docs don't specify but I dove through the source to make sure this would work 😂

src/components/code-block/index.js
developit: I'll check - I think I was seeing reapplication of HTML here, but you're right there is no reason Preact would be doing that. Was likely something else and I can un-patch.

PR #380, opened 2019-08-08
Markdown worker

developit:
Depends on #379 (retarget once merged).

Note: I split this out because I'm not sure if it's better or worse for performance. This PR will get a Netlify preview that we can use to compare against the one in #379 via webpagetest.org/easy.

### DIFF
--- a/src/components/content-region.js
+++ b/src/components/content-region.js
@@ -1,4 +1,4 @@
-import Markdown from './markdown';
+import Markup from 'preact-markup';
 import widgets from './widgets';
 
 const COMPONENTS = {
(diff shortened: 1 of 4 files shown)


developit:
weird.. no travis?

marvinhagemeister:
@developit The CI only runs when the target branch is `master` :+1:

developit:
Ah, I'm silly

marvinhagemeister:
Nah, don't say that. It happens to everyone :slightly_smiling_face:

developit:
Lighthouse perf score: 97
https://www.webpagetest.org/result/190808_A6_84da7676755460c6103a200465af2c54/

PR #381, opened 2019-08-08
Upgrade REPL

developit:
Depends on #379 (retarget once merged).


src/components/code-block/index.js
```diff
@@ -68,6 +72,8 @@ function HighlightedCodeBlock({ code, lang, ...props }) {
 		code.replace(/</g, '&lt;').replace(/>/g, '&gt;');
 	const htmlObj = useMemo(() => ({ __html: html }), [html]);
 
+	console.log({ canHighlight, pending, error });
```
marvinhagemeister: whoops :slightly_smiling_face:

src/components/controllers/repl/index.js
```diff
@@ -108,6 +108,7 @@ export default class Repl extends Component {
 	}
 
 	render(_, { loading, code, error, example, copied }) {
+		console.log('error: ', error);
```
marvinhagemeister: Let's only log when there is an error :+1:

src/components/controllers/repl/runner.js
```diff
@@ -99,7 +104,7 @@ export default class Runner extends Component {
 		}
 		if (vnode) {
 			try {
-				this.root = render(vnode, this.base, this.root);
+				this.root = render(vnode, this.base);
```
marvinhagemeister: I wonder if we should wrap a component with a `componentDidCatch` lifecycle all around the users code.

developit: ooh I like that idea!

PR #383, opened 2019-08-08
Basic REPL error handling

marvinhagemeister:
This is a ~~super super hacky~~ draft of error handling for the Repl and obviously needs to be refactored. It's more of a starting point to get to know what CodeMirror expects to properly display the error.

- [x] Add pretty error overlay over preview area
  - [x] ~~Clickable stacktrace to jump into editor?~~ - let's keep it simple for now
- [x] Show inline Error in CodeMirror

Open Questions:

1) Should we move the `Runner` to an `iframe`? The current way to handle uncaught `Promise` exceptions is a bit meh.

localhost_8080_repl (1)


developit:
We'd have to decide whether we want to use an on-domain frame or not. Maybe on-domain with its own copy of Preact? That would help avoid the potential issue where someone writes REPL code that pushes an unrecoverable component error into the state update queue so the site deadlocks. Though that could also potentially be solved by wrapping the root in a componentDidCatch as you mentioned.

developit:
For the error stack trace, I can turn on line-accurate sourcemaps from sucrase. Didn't have time in my initial PR.

PR #391, opened 2019-08-12
Add section about custom events

marvinhagemeister:
This PR adds a little section about custom events on web components. See [this discussion](https://github.com/preactjs/preact/issues/1820) for the reason for this change.

### DIFF
--- a/content/en/guide/v10/web-components.md
+++ b/content/en/guide/v10/web-components.md
@@ -45,7 +45,7 @@ function Foo() {
 
 ## Triggering custom events
 
-A web component might listen for custom events to add additional functionality. We can't make the same assumptions about casing like we can with native DOM events, so we pass the event name as is. We don't change the casing in any way there.
+Preact normalizes the casing of standard built-in DOM Events, which is how we can pass an `onChange` prop to `<div>`, when the event listener actually requires lowercase `"change"`. However, Custom Elements often fire custom events as part of their public API, and there's no way to know what custom events might be fired. In order to ensure Custom Elements are seamlessly supported in Preact, any unrecognized event handler props passed to a DOM Element will have their casing preserved.
 
 ```jsx
 // native DOM event -> listens for a "click" event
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/web-components.md
```diff
@@ -42,3 +42,18 @@ function Foo() {
   return <x-foo ref={myRef} />;
 }
 ```
+
+## Triggering custom events
+
+A web component might listen for custom events to add additional functionality. We can't make the same assumptions about casing like we can with native DOM events, so we pass the event name as is. We don't change the casing in any way there.
```
developit:
```suggestion
Preact normalizes the casing of standard built-in DOM Events, which is how we can pass an `onChange` prop to `<div>`, when the event listener actually requires lowercase `"change"`. However, Custom Elements often fire custom events as part of their public API, and there's no way to know what custom events might be fired. In order to ensure Custom Elements are seamlessly supported in Preact, any unrecognized event handler props passed to a DOM Element will have their casing preserved.
```

content/en/guide/v10/web-components.md
```diff
@@ -42,3 +42,18 @@ function Foo() {
...
+
+```jsx
+// native DOM event -> add handler for "click" event
```
developit:
```suggestion
// native DOM event -> listens for a "click" event
```

content/en/guide/v10/web-components.md
```diff
@@ -42,3 +42,18 @@ function Foo() {
...
+<div onClick={() => console.log('click')} />
+
+// Custom Element
+// Add handler for "IonChange" event
+<my-foo onIonChange={() => console.log('IonChange')} />
+// Add handler for "ionChange" event (note the casing)
+<my-foo onionChange={() => console.log('IonChange')} />
```
developit:
```suggestion
<my-foo onionChange={() => console.log('ionChange')} />
```

PR #393, opened 2019-08-12
WIP fix for upgrading from the old site SW

developit:
Upgrading from the old site for those who've previously visited it currently requires a refresh, since the `sw.js` did not use `skipWaiting()`.


prateekbh:
Need to think how will the modern browser users upgrade from `sw.js` to `sw-esm.js`

prateekbh:
Turns out its takes care of by the browser 👍

src/sw-fix.js
```diff
@@ -0,0 +1,90 @@
+// this gets appended to the built sw
+
+/*global clients*/
+
+// eslint-disable-next-line
+var storage = (function(e) {
+	'use strict';
```
prateekbh: Travis is unhappy with this

developit: hmm - this was previously all on one line, which meant the ESLint ignore on L5 applied to the whole thing.

developit:
(fixed elsewhere!)

PR #400, opened 2019-08-14
Fix h/2 push header duplication causing duplicate pushes

developit:


developit:
I really wish there was a way to unset a header in `_headers` haha

calebeby:
BTW Netlify does not support H2 push rn (they disabled it and deleted the blog post announcing it).

https://community.netlify.com/t/conditional-headers-based-on-user-agent/415/2?u=calebeby

calebeby:
So what you are doing is only preload, it doesn't trigger push

PR #401, opened 2019-08-14
fix occasional test timeout

developit:


test/browser/homepage.test.js
```diff
@@ -95,6 +95,9 @@ describe('Homepage', () => {
 
 		await click('a[href="/guide/v10/getting-started"]');
 
+		// sleep 1s
+		await new Promise(r => setTimeout(r, 1000));
```
marvinhagemeister: I wonder if instead of using a fixed timeout we should use [waitForSelector](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitforselectorselector-options). Something like:

```js
await page.waitForXPath("//h1[contains(., 'Expected text')]")
```

rschristian:
Closing, tests were removed at some point

PR #402, opened 2019-08-14
Attempt to achieve prefetch for workers

developit:


package.json
```diff
@@ -11,7 +11,7 @@
     "start": "npm run -s server",
     "prestart": "npm run -s build",
     "server": "superstatic build -p ${PORT:-8080} --host 0.0.0.0 --gzip -c '{\"rewrites\": [{\"source\":\"**\",\"destination\":\"index.html\"}],\"headers\":[{\"source\":\"**\",\"headers\":[{\"key\":\"Cache-Control\",\"value\":\"max-age=31536000\"}]}]}'",
-    "build": "preact build --prerenderUrls src/prerender.js --template src/template.html",
+    "build": "preact build --no-esm --prerenderUrls src/prerender.js --template src/template.html",
     "postbuild": "cat src/_headers >> build/_headers",
```
marvinhagemeister: No `esm` build?

developit: it sortof breaks with the workers, since they never have .esm.js

PR #413, opened 2019-08-29
Correction of the word Performante by Performance

yhau1989:
Correction of the word Performante by Performance


rschristian (approved):
Well, hopefully late is better than never 😅

PR #416, opened 2019-08-30
(chore) Update upgrade guide for preact-redux

cristianbote:
I've marked it as draft until `preact-redux@3.0.0` will be released.


rschristian:
Going to close this out as it looks like the recommendation is to use `react-redux`

PR #419, opened 2019-09-02
Rewrite compat page

marvinhagemeister:
This PR rewrites the page about `preact/compat` or otherwise known as `Switching to Preact`. That was the last page which still referenced `preact-compat`.

### DIFF
--- a/content/en/guide/v10/differences-to-react.md
+++ b/content/en/guide/v10/differences-to-react.md
@@ -143,7 +143,7 @@ function App(props) {
 [preact/compat] ships with specialised components that are not necessary for every app. These include
 
 - [PureComponent](/guide/v10/switching-to-preact#purecomponent): Only updates if `props` and `state` have changed
-- [memo](/guide/v10/switching-to-preact#memo): Similar in spirit to `PureComponent` but allows to use a custom comparer function
+- [memo](/guide/v10/switching-to-preact#memo): Similar in spirit to `PureComponent` but allows to use a custom comparison function
 - [forwardRef](/guide/v10/switching-to-preact#forwardRef): Supply a `ref` to a specified child component.
 - [Portals](/guide/v10/switching-to-preact#portals): Continues rendering the current tree into a different DOM container
 - [Suspense](/guide/v10/switching-to-preact#suspense): **experimental** Allows to display fallback content in case the tree is not ready
--- a/content/en/guide/v10/differences-to-react.md
+++ b/content/en/guide/v10/differences-to-react.md
@@ -142,14 +142,14 @@ function App(props) {
 
 [preact/compat] ships with specialised components that are not necessary for every app. These include
 
-- `PureComponent`: Only updates if `props` and `state` have changed
-- `memo`: Similar in spirit to `PureComponent` but allows to use a custom comparer function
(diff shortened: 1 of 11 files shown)


content/en/guide/v10/differences-to-react.md
```diff
@@ -142,14 +142,14 @@ function App(props) {
...
 [preact/compat] ships with specialised components that are not necessary for every app. These include
 
-- `PureComponent`: Only updates if `props` and `state` have changed
-- `memo`: Similar in spirit to `PureComponent` but allows to use a custom comparer function
-- `forwardRef`: Supply a `ref` to a specified child component.
-- `Portals`: Continues rendering the current tree into a different DOM container
-- `Suspense`: **experimental** Allows to display fallback content in case the tree is not ready
-- `lazy`: **experimental** Lazy load async code and mark a tree as ready/not ready accordingly.
+- [PureComponent](/guide/v10/switching-to-preact#purecomponent): Only updates if `props` and `state` have changed
+- [memo](/guide/v10/switching-to-preact#memo): Similar in spirit to `PureComponent` but allows to use a custom comparer function
```
jviide: "Comparison function" seems to be used in the compat docs as well as the React docs (instead of "comparer function").

content/en/guide/v10/switching-to-preact.md
```diff
@@ -6,290 +6,137 @@ description: 'Everything you need to know to switch from React to Preact.'
...
-    "react": "preact-compat",
-    "react-dom": "preact-compat"
-  }
-}
+// Render a second time, doesn't log anything
+render(<Foo value="3" />, dom);
 ```
 
-#### Aliasing via Browserify
+> Note that the advantage of `PureComponent` only pays off when then render is expensive. For simple children trees it can be quicker to just do the render compared to the overhead of comparing props.
```
jviide: render -> `render`?

content/en/guide/v10/switching-to-preact.md
```diff
@@ -6,290 +6,137 @@ description: 'Everything you need to know to switch from React to Preact.'
...
 
-If you're using Browserify, aliases can be defined by adding the [aliasify](https://www.npmjs.com/package/aliasify) transform.
+## memo
 
-First, install the transform:  `npm i -D aliasify`
+`memo` is equivalen to functional components as `PureComponent` is to classes. It uses the same comparison function under the hood, but allows you to specify your own specialized function optimized for your use case.
```
jviide: equivalen -> equivalent

content/en/guide/v10/switching-to-preact.md
```diff
@@ -6,290 +6,137 @@ description: 'Everything you need to know to switch from React to Preact.'
...
 
-```js
-// initial render:
-render(<App />, document.body);
+## Suspense (experimental)
 
-// update in-place:
-render(<App />, document.body, document.body.lastElementChild);
-```
+The main idea behind `Suspense` is to allow sections of your UI to display some sort of loading content while components further down the tree are still loading. A common use case for this is code-splitting where you'll need to load a component from the network before you can render it.
```
jviide: Maybe "placeholder content" instead of "loading content"?

PR #422, opened 2019-09-03
add polyfills to workers

developit:


developit:
Just a note - this still doesn't have the REPL working in IE11, but everything else works. I might do a followup PR to remove `@babel/polyfill` for now since it's not fixing the REPL anyway.

marvinhagemeister:
Yeah I merged it, because it looked to be ready :sweat_smile:

PR #441, opened 2019-10-01
Add missing word in tutorial.md

Otto-AA:
Not 100% sure if this is the correct tense, but the you is definitely missing there


ForsakenHarmony:
Yup looks like it, thanks for the PR

PR #447, opened 2019-10-02
Remove "next" tags from docs

andrewiggins:
Now that Preact X is released, users no longer need to use the "next" tag to install it :tada:

Note:
Before we complete this PR, the "latest" tag for `preact-jsx-chai` and `preact-markup` tag need to be moved up to the appropriate Preact X versions


marvinhagemeister:
Afaik the other languages haven't really been updated for X yet. I'm happy to merge this once `preact-jsx-chai` and `preact-markup` are updated 👍

PR #450, opened 2019-10-03
Fix typo in CSS Custom Properties example

bingocaller:
Div tag was closed prematurely, which also resulted in a non-functioning
example, when run in REPL.


bingocaller:
No problem! Thank _**you**_ and the rest of the team for all your hard work! Preact is amazing and I can't wait to build my first Preact X project! 🙌

PR #451, opened 2019-10-05
Fix code sample

guido4000:
?module expands all “bare” import specifiers in JavaScript modules to unpkg URLs. This is necessary. Otherwise the imports can not be found.


developit:
@guido4000 there are no import specifiers in either of those modules because they have no dependencies. 

This is required for `unpkg.com/htm/preact` though. I think there was a broken code sample that I fixed a couple days ago for that.

guido4000:
That is strange.

When I remove the module parameter from the htm import I get the following error in Chrome: 
Uncaught SyntaxError: The requested module 'https://unpkg.com/htm' does not provide an export named 'default'

When I remove the module parameter from the preact import I get this error in the console:
Uncaught SyntaxError: The requested module 'https://unpkg.com/preact' does not provide an export named 'h'

developit:
Ah, sorry @guido4000 you're right. Here `?module` is just triggering unpkg to use the `"module"` package.json field rather than `"main"`, which is desirable.

PR #456, opened 2019-10-12
Remove "Remove the 3rd argument to render"

38elements:
I may be misunderstanding `render()`.
I think `replaceNode` parameter is returned by this pull request.
https://github.com/preactjs/preact/pull/1557


marvinhagemeister:
That's true, the third argument came back in a later release. I'd still love to keep some of the text because the common case can be done without it. This is different from Preact 8 where it was required to have React's behaviour :+1:

38elements:
@marvinhagemeister 
Thank you for reply.
IMHO, I think it is better that appending and replacing is indicated expressly.

PR #466, opened 2019-10-28
translate pt-br

devalexandre:


content/de/about/we-are-using.md
```diff
@@ -280,6 +280,11 @@ permalink: '/about/we-are-using'
         </a>
     </li>
     <li>
+
+        <a href="//www.indev.net.br" target="_blank">
+            <img src="/assets/we-are-using/indevweb.png" alt="Indev Desenvolvimento web">
```
marvinhagemeister: I think adding a new company is not necessary to translate our docs. I'd rather have it in a separate PR.

devalexandre:
why is blocked ?

content/de/about/we-are-using.md
```diff
@@ -1,314 +0,0 @@
----
-name: We Are Using
-title: Wer benutzt Preact?
```
marvinhagemeister: This doesn't look correct. This PR deletes the `We are using` page for every language. Maybe this change got included in this PR by accident? It's also the reason why the CI is failing.

PR #469, opened 2019-11-01
Manually filter recent version by semver from github releases li…

marvinhagemeister:
~~I'd love to have it dynamic, but to do that we'd need to query npm instead. GitHub doesn't support multiple release lines. So it's better to show the correct version statically than to show a wrong one.~~

Because GitHub doesn't support multiple release lines we would previously always pull the latest one by date. So in effect our version number would switch between the `8.x` and `10.x` release line. That's not what we want!

By suggestion from @cristianbote , this PR now changes that to load the release list and we'll find the most recent one ourselves from that.

### DIFF
--- a/src/components/header/gh-version.js
+++ b/src/components/header/gh-version.js
@@ -1,12 +1,30 @@
 import { h } from 'preact';
+import { useEffect, useState } from 'preact/hooks';
+import { fetchRelease } from '../../lib/github';
+import config from '../../config.json';
 
 const VERSION = '10.0.4';
 const URL = `https://github.com/preactjs/preact/releases/tag/${VERSION}`;
 
 export default function ReleaseLink(props) {
+	const [url, setUrl] = useState(URL);
+	const [version, setVersion] = useState(VERSION);
+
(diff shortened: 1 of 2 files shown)


src/components/header/gh-version.js
```diff
@@ -1,36 +1,12 @@
...
-
-let VERSION = '10.0.0';
-if (PRERENDER) {
-	VERSION = require('../../../package.json').dependencies.preact.replace(
-		'^',
-		''
-	);
-}
+const VERSION = '10.0.4';
+const URL = `https://github.com/preactjs/preact/releases/tag/${VERSION}`;
```
cristianbote: Could we fetch all the tags and sort them based on semver rather than latest?

marvinhagemeister: That would work too. I have a branch ready for that, but I'm running into rate limiting issues with the API. Looks like there is a bit more work involved here for a proper longterm solution.

marvinhagemeister: Nevermind, I added it.

PR #487, opened 2019-11-24
Feature language select

vitormalencar:


ForsakenHarmony:
It's a bit cramped up there in the header, maybe it should show the country flag for the language

vitormalencar:
Yep I have the same idea, I think it would be better, for the 1 PR I went for something very simple, maybe I should for the flags right the way what you think @ForsakenHarmony?

marvinhagemeister:
Agree with @ForsakenHarmony, I think we have too many elements in the header.

Screenshot from 2019-11-26 13-05-59

Not sure what a better solution would look like. Throwing out some random ideas:

1. What if we introduce a `settings` icon. We could move both the language dropdown and the theme toggle to that. A disadvantage would be that it's hidden again behind something.

vitormalencar:
I like this idea @marvinhagemeister is not one lick away but it fixes the problem of scrolling to the bottom =),
also, another idea is to do like the react docs, and create a link for a language page that we could list all the translations.

Screen Shot 2019-11-26 at 12 41 16

what you think @ForsakenHarmony and @marvinhagemeister good or over complicated?

marvinhagemeister:
A language page is also a great option! I'm not sure if we'll ever have as many translations as React but you never know!

PTortello (approved):

vitormalencar:
thanks for the Approval @PTortello do we think that this feature is still wanted ?

ForsakenHarmony:
I think this has been implemented, we have a selector in the footer

PR #494, opened 2019-11-27
Add SwissDev Jobs logo

Vrq:


Vrq:
No probs, I also missed it in the previous one :)

Do you know when it gets deployed, because I still cannot see the logo?

marvinhagemeister:
@Vrq we deploy automatically when something is merged into `master`. A quick reload should fetch the new asset :+1:

PR #497, opened 2019-12-07
Remove cached content if Promise rejected

marvinhagemeister:


rschristian:
Cleaning up old PRs, going to close this out as we've vastly changed things since and this is no longer relevant.

PR #500, opened 2019-12-09
Make the mobile sidebar/ToC more visible and obvious

developit:


marvinhagemeister:
Awesome, I really love this design 🙌

We should add a bit more padding to the bottom of the mobile sidebar menu, so that the last item isn't hidden by the guide button.
Screenshot_20191210-110603

developit:
Agreed @marvinhagemeister. I was originally going to have the button move to the right, but that felt a little unfair to left-handed folks and made toggling the menu quickly fairly difficult. Also it required animating the `left` property, which I'm averse to doing 🤣.

PR #509, opened 2020-01-02
Add twitter card

38elements:
This adds [Twitter card](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started).

https://deploy-preview-509--preactjs.netlify.com/guide/v10/getting-started

You can check by this page.
https://cards-dev.twitter.com/validator


developit:
Nice! We should find a way to populate these from prerender-urls.

PR #515, opened 2020-01-07
【preact/compat】Update switching-to-preact.md

YikaJ:
in preact v10, need to update to preact/compat


marvinhagemeister:
For Preact 8.x it's correct to use `preact-compat` instead of `preact/compat`. The latter is only required starting with 10.x and onwards.

PR #539, opened 2020-02-12
Use `onChange` in select example

andrewiggins:
Update the `select` example to use the `onChange` event which has better browser support (namely IE & old-Edge)


content/en/guide/v10/forms.md
```diff
@@ -83,7 +83,7 @@ class MySelect extends Component {
   render(_, { value }) {
     return (
       <form onSubmit={this.onSubmit}>
-        <select value={value} onInput={this.onInput}>
+        <select value={value} onChange={this.onInput}>
```
andrewiggins: Should I change the name of the method as well?

developit: yeah, lets.

andrewiggins: Done

andrewiggins:
I also updated the example to display the the current value in the alert box so its easier to verify that the sample works in the REPL

PR #549, opened 2020-02-29
Add Gold and Platinum sponsors to home page

marvinhagemeister:
Per our tiers on [OpenCollective](https://opencollective.com/preact) we need to display gold + platinum sponsors on our home page. This PR addresses that.

Dark:

Screenshot from 2020-02-29 13-37-02

Bright:
Screenshot from 2020-02-29 13-54-45


src/components/sponsors/index.js
```diff
@@ -0,0 +1,79 @@
...
+					<a href="https://webflow.com/" title="webflow">
+						<WebflowLogo />
+					</a>
+				</li>
+			</ul>
+		</div>
+	);
+}
+
+function AMPLogo() {
```
developit: Any chance we could use `` rather than inline SVG here? It'd keep DOM size down and bundles small.

marvinhagemeister: Sure thing, switched to external svg files with the next commit :+1:

PR #552, opened 2020-02-29
Add categories to the sidebar

marvinhagemeister:
Introduce categories to the sidebar on the docs page to make them easier scanable. The long-term goal is to have more pages which explain the concepts and ideas behind certain features in Preact more in detail instead of being more or less just an API reference.

Some form of grouping should it also make easier to grasp for newcomers which topics are important and can potentially be postponed to a later learning stage.

Tried various variants for the accordion and most other doc pages seems to allow only one to be open at a time. I've played with that, but imo it doesn't feel right as ctrl+f won't find the hidden text.

Screenshot from 2020-02-29 22-12-33


developit:
I *LOVE* this.

developit:
I tried thinking about any slight changes in categories, here's an option I came up with:

```
INTRODUCTION
  Getting Started
  What's new?
  Upgrading from Preact 8
  Tutorial
ESSENTIALS
  Components
  Hooks
  Forms
  References
  Context
DEBUG & TEST
  Debugging Tools
  Unit Testing with Enzyme
REACT COMPATIBILITY
  Differences to React
  Switching to Preact
ADVANCED
  API Reference
  Web Components
  Server-Side Rendering
  Progressive Web Apps
  External DOM Mutations
  Extending Component  (delete?)
  Options
```

marvinhagemeister:
Good idea! Changed the ordering like in your snippet. Agree about the `Extending Components` page. It's pretty outdated and can be deleted :+1:

PR #556, opened 2020-03-02
reminder to call the script as a module

areai51:


developit:
Hi @areai51! You make a great point about this. I've actually gone ahead and modified this PR to make some more extensive changes to this part of our documentation - let me know what you think! I always thought it would be clearer what was going on if we provided the full HTML for these examples rather than just the JavaScript portion. 

@marvinhagemeister I'll probably swap these "edit on glitch" links back to our own REPL, but I need to add support for unwrapping `<script>` in `html` examples first.

areai51:
@developit  yup I agree providing the complete source including the HTML makes it a lot more easier for people to understand.

developit:
Awesome! Sorry for bloating this PR! haha

developit:
woot!

PR #562, opened 2020-03-04
dark mode updates for company logos

developit:
this is my attempt to make the black-colored logos visible on hover. it's not perfect, but at least they show up.


developit:
superceded by #563

PR #584, opened 2020-04-18
Document selective context

JoviDeCroock:
Depends on https://github.com/preactjs/preact/pull/2452

### DIFF
--- a/content/en/guide/v10/hooks.md
+++ b/content/en/guide/v10/hooks.md
@@ -282,7 +282,7 @@ The `useContext` hook accepts an optional `shouldUpdate` function as a second ar
 const Theme = createContext({ theme: 'light' });
 
 function DisplayTheme() {
-  const theme = useContext(Theme, (ctx, prevCtx) => ctx.theme !== prevCtx.theme);
+  const { theme } = useContext(Theme, (ctx, prevCtx) => ctx.theme !== prevCtx.theme);
   return <p>Active theme: {theme}</p>;
 }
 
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/hooks.md
```diff
@@ -276,6 +276,33 @@ function App() {
 }
 ```
 
+The `useContext` hook accepts a second arguments, we call this the `shouldUpdate` function, this will be registered once and will determine wheter or not your component should rerender.
```
developit, line 279:
```suggestion
The `useContext` hook accepts an optional `shouldUpdate` function as a second argument. This function will be called each time the context value is changed, and can return a Boolean value indicating whether the subscribed component should rerender.
```

content/en/guide/v10/hooks.md
```diff
@@ -276,6 +276,33 @@ function App() {
...
+
+```jsx
+const Theme = createContext({ theme: 'light' });
+
+function DisplayTheme() {
+  const theme = useContext(Theme, (ctx, prevCtx) => ctx.theme !== prevCtx.theme);
```
developit, line 285:
```suggestion
  const { theme } = useContext(Theme, (ctx, prevCtx) => ctx.theme !== prevCtx.theme);
```

content/en/guide/v10/hooks.md
```diff
@@ -276,6 +276,33 @@ function App() {
...
+    <Theme.Provider value={{ theme: 'light' }}>
+      <OtherComponent>
+        <DisplayTheme />
+      </OtherComponent>
+    </Theme.Provider>
+  )
+}
+```
+
+In the above example we'll only rerender if the current context and previous context are different from one another.
```
developit, line 301:
```suggestion
In the above example, the custom `shouldUpdate` function means `DisplayTheme` will only be re-rendered when the context object's `.theme` property is changed.
```

content/en/guide/v10/hooks.md
```diff
@@ -276,6 +276,33 @@ function App() {
...
+
+> Note that using a conditional there like `condition ? func1 : func2` will only use the function from the first value since it doesn't
+  track the need for re-registering.
```
developit, line 304:
```suggestion
> Note: the `shouldUpdate` function is registered once and can't be changed dynamically. New function references passed as the second argument to useContext will be ignored.
```

PR #585, opened 2020-04-23
tweaks

38elements:
It seems that `content/en/about/cli.md` is not used.


content/en/index.md
```diff
@@ -136,7 +136,7 @@ function Counter() {
 
 <section class="home-split">
     <div>
-        <h2>Todo List Component</h2>
+        <h2>Todo List</h2>
```
38elements, line 139: `Fetch GitHub Stars` is not `Fetch GitHub Stars Component`.
Therefore, This  is consistent.

PR #591, opened 2020-05-16
Improve content/en/guide/v10/forms.md

38elements:
This improves content/en/guide/v10/forms.md.


content/en/guide/v10/forms.md
```diff
@@ -5,37 +5,39 @@ description: 'How to build awesome forms in Preact that work anywhere.'
 
 # Forms
 
-Forms in Preact work much the same as they do in HTML. You render a control, and attach an event listener to it.
```
developit, line 8: hey @38elements - I tried to rephrase this so that the title ends up being a single line. Let me know what you think!

content/en/guide/v10/forms.md
```diff
@@ -95,20 +97,39 @@ class MySelect extends Component {
...
+          onClick={this.toggle}
+        />
+      </label>
+    );
+  }
+}
+```
 
-### Checkbox Example
+But, it is unnecessary to use `checked` value in Event object, since the component has `checked` value at before changing in `state`. So, instead of listening for a `change` event we should listen for a `click` event, which is fired any time the user clicks on the checkbox _or an associated `<label>`_. Checkboxes just toggle between Boolean `true` and `false`, so clicking the checkbox or the label, we'll just invert whatever value we have in state, triggering a re-render, setting the checkbox's displayed value to the one we want.
```
developit, line 127:
```suggestion
When managing the state of checkboxes or radio buttons, it's important to understand the two different mechanisms for observing value changes. The `click` event is triggered on an input when its own value changes, such as when a radio button is clicked, or activated via the keyboard or its associated `<label>`. The `change` event is triggered on all inputs in a form group when the value of that _group_ is changed.

It's most common to use the `click` event for checkboxes and radio buttons, since it doesn't rely on named form groups which are a type of DOM state. When handling `click` events, the input element's boolean `checked` property indicates its current value.

In the example below, clicking on the checkbox or its label should toggle between Boolean `true` and `false` values. To achieve this as a controlled input, we can store the initial value in state and simply invert it in response to clicks. This will trigger a re-render, setting the checkbox's displayed value to the value from state.
```

content/en/guide/v10/forms.md
38elements: Thank you for review.

I think "Forms in Preact are largely the same as HTML forms.
The main difference between both is that in most cases the value of input control is controlled by whether DOM node or Component.	Much like in HTML forms, forms in Preact both work by rendering input elements and listening for events triggered as the form is modified. " is better.

I think "The main difference is that in most cases the input controll's value is not controlled by the DOM node, but by component." is better. 

"most cases the value" => "most cases the input controll's value"
"but by Preact" => "but by component"

This explicitly describes the difference in value management between DOM node and component. I think this easily lead to the next sentence.

content/en/guide/v10/forms.md
```diff
@@ -95,20 +97,39 @@ class MySelect extends Component {
...
+Say we listen for a "change" event on a checkbox in Controlled Component, which is fired when the checkbox is checked or unchecked by the user. Form component set `checked` value in the Event object received from the checkbox to `state` in `change` event handler. This triggers rerendering component. `checked` value in `state` is reflected in the checkbox by this rerendering.
 
-> **Walk-Through:** Say we listen for a "change" event on a checkbox, which is fired when the checkbox is checked or unchecked by the user.  In our change event handler, we set a value in `state` to the new value received from the checkbox.  Doing so will trigger a re-render of our component, which will re-assign the value of the checkbox to the value from state.  This is unnecessary, because we just asked the DOM for a value but then told it to render again with whatever value we wanted.
```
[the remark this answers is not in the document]
38elements: Does " This is unnecessary, because we just asked the DOM for a value but then told it to render again with whatever value we wanted." mean the following?

```jsx
class MyForm extends Component {
  toggle = event => {
      const checked = event.target.checked;
      this.setState({ checked });
  };
  render(_, { checked }) {
    return (
      <label>
        <input
          type="checkbox"
          onChange={this.toggle}
        />
        foo
      </label>
    );
  }
}
```

I may be misunderstanding.

content/en/guide/v10/forms.md
developit: There are some grammar issues in the first one, but I agree it's worth comparing to the DOM version.
What about this:

>Forms in Preact are largely the same as HTML forms.
>
> Like in HTML, forms in Preact work by rendering input elements and listening for events triggered as the form is modified. The main difference when rendering with Preact is that the value of a form input is generally controlled by a Component, rather than by the DOM element itself.

38elements: I think your suggestion is better.

content/en/guide/v10/forms.md
```diff
@@ -5,37 +5,46 @@ description: 'How to build awesome forms in Preact that work anywhere.'
...
 
 <div><toc></toc></div>
 
 ---
 
-## Controlled & Uncontrolled Components
+## Controlled Component and Uncontrolled Component
 
-When talking about form controls you'll often encounter the words "Controlled Component" and "Uncontrolled Component". The description refers to the way data flow is handled. The DOM has a bidirectional data flow, because every form control will manage the user input themselves. A simple text input will always update it's value when a user typed into it.
+You'll often encounter the terms "Controlled Component" and "Uncontrolled Component" in documentation about working with forms in Preact. These terms refer to the way data flow is handled. The DOM has bidirectional data flow where each form control manages its own state and user input. For example, when typing into a text input element, the text is updated without requiring additional logic to store and display values.
```
38elements, line 20: Considering consistency with following, I think `"Controlled" Component and "Uncontrolled" Component` is better than `"Controlled Component" and "Uncontrolled Component"`.  I think "Controlled" and  "Uncontrolled" refer to the way data flow is handled.


> A "Controlled" Component is a component where state must be manually managed by listening for events and passing new values as props. An "Uncontrolled" Component is a component that manages its own internal state, handling events and updating itself automatically.

content/en/guide/v10/forms.md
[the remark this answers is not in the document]
38elements: This sentence is difficult for me.
I think "we just asked the DOM for a value" is "getting checkbox's value".
What exactly does "whatever value we wanted" mean?
"This is unnecessary, because we just asked the DOM for a value but then told it to render again with whatever value we wanted."

Issue #615, opened 2020-06-16
Add docs about SVG normalization


teodragovic:
Going over issues in main repo (https://github.com/preactjs/preact/issues/2404, https://github.com/preactjs/preact/issues/1061, https://github.com/preactjs/preact/issues/813, https://github.com/preactjs/preact/issues/1940), misunderstanding about SVG props normalization seems common enough to be added into "Differences to React" section (note about following DOM specs might imply this but it's not very obvious).

IMO @andrewiggins already provided great explanation [here](https://github.com/preactjs/preact/issues/1940#issuecomment-548703893) that can be used as basis.

I would gladly submit PR if that's ok?

developit:
@teodragovic It seems like most of those issues were fixed. Is the idea that this would go under "Differences to React" with an explanation that while `preact` does not perform normalization, `preact/compat` addresses the differences?

teodragovic:
Yep, exactly.

marvinhagemeister:
sounds good to me. We should point out that Preact itself follows the various HTML specs in naming and that only compat normalizes those like in React.

PR #617, opened 2020-06-22
Add note about passing SVG attributes through

teodragovic:
Fixes https://github.com/preactjs/preact-www/issues/615


content/en/guide/v10/differences-to-react.md
```diff
@@ -85,6 +85,26 @@ standard `class` attribute instead of `className`.
 
 Most Preact developers prefer to use `class` because it's shorter to write, but both are supported.
 
+### Raw SVG inside JSX
+
+SVG is pretty interesting when it comes to the names of its properties and attributes. Some properties (and their attributes) on SVG objects are camelCased (e.g. [clipPathUnits on a clipPath element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/clipPath#Attributes)), some attributes are kebab-case (e.g. [clip-path on many SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/Presentation)), and other attributes (usually ones inherited from the DOM, e.g. `oninput`) are all lowercase.
+
+Preact doesn't normalize SVG properties and attributes [like React does]((https://github.com/facebook/react/pull/6243)) and instead, allows you to write valid SVG inside JSX by passing attributes through.
```
marvinhagemeister, line 92: What are your thoughts about rephrasing that section to show the advantages of Preact's approach? Something akin to:

```
Preact forwards SVG-Attributes as is. This allows you to copy and paste unmodified SVG snippets right into your code and have them work out of the box. This allows greater interoperability with tools designers tend to use to generate icons or SVG illustrations.

If you're coming from React you're likely used to specify every attribute in camelCase. If you'd like to continue using the camelCase'd attribute names you can use our [preact/compat] compatibility layer. It mirrors the React API and normalizes these attributes.
```

What do you think?

content/en/guide/v10/differences-to-react.md
```diff
@@ -85,6 +85,26 @@ standard `class` attribute instead of `className`.
 
 Most Preact developers prefer to use `class` because it's shorter to write, but both are supported.
 
+### Raw SVG inside JSX
```
marvinhagemeister, line 88: I'd remove the "Raw" here. Not sure what an "un-raw" SVG would look like.

content/en/guide/v10/differences-to-react.md
teodragovic: Yes! My initial intent was to phrase it as a feature more than un-react thing. Should I drop the first paragraph about SVG? I realize it's a bit broad and out of scope. I copied it verbatim since it was very useful when I was investigating this particular issue.

marvinhagemeister: Noo, I love the first paragraph here! It makes the doc sound way more "human" 🙌 I'd just change the second one as it can be interpreted wrongly when potential users read "Preact doesn't normalize SVG properties and attributes" without there being an explanation why we made that choice.

teodragovic: done!

developit:
This is really great @teodragovic! Sorry for the late review.

PR #619, opened 2020-06-25
Added new company in the list

Amalkarunakaran:
added Extrieve Technologies in the list

### DIFF
--- a/src/components/we-are-using/index.js
+++ b/src/components/we-are-using/index.js
@@ -247,6 +247,10 @@ const companies = [
 		name: 'Aarogya Setu',
 		href: 'https://www.aarogyasetu.gov.in/',
 		logo: 'aarogyasetu.svg'
+	},{
+		name: 'Extrieve Technologies',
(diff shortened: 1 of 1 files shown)


src/components/we-are-using/index.js
```diff
@@ -247,6 +247,10 @@ const companies = [
 		name: 'Aarogya Setu',
 		href: 'https://www.aarogyasetu.gov.in/',
 		logo: 'aarogyasetu.svg'
+	},{
+		name: 'Extrieve Technologies',
+		href: 'https://www.extrieve.com/',
```
marvinhagemeister, line 252: Something is off with the SSL certificate:

Screenshot 2020-06-25 at 14 21 21

reznord: Loads only via `http`

reznord:
This site is not using preact at all. Checked in the browser console. 

So, closing this for now.

Feel free to re-open when the site uses preact :)

PR #622, opened 2020-07-01
Removing border from logos.

Azuriru:
Might I suggest killing that out of place .5px drop shadow. It looks really bad.


developit:
The problem is that some of the logos are jet black and basically invisible without the border:

PR #628, opened 2020-07-13
fix link to #usecontext hook

mgrubinger:
The link to the useContext hook currently is set to `#context`, but should be `#usecontext`.

I also noticed that routing does not handle anchors correctly, i.e. the link `https://preactjs.com/guide/v10/hooks#usecontext` works when the page is initially loaded, but does not jump to the anchor after client side navigation happened (tested in Chrome and Firefox). Do you want me to create a ticket for that?

### DIFF
--- a/content/en/guide/v10/context.md
+++ b/content/en/guide/v10/context.md
@@ -43,7 +43,7 @@ function App() {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
@mgrubinger That sounds like an something that's worth to file an issue for :+1:

PR #629, opened 2020-07-13
move to a netlify function for GH interactions

JoviDeCroock:


src/lambda/repo.js
```diff
@@ -0,0 +1,34 @@
...
+	const CACHE = {};
+	return key => CACHE[key] || (CACHE[key] = fn(key));
+}
+
+// TODO: use GraphQL to avoid fetching so much data
+const repoInfo = memoize(repo =>
+	fetch(`https://api.github.com/repos/${repo}`)
+		.then(checkStatus)
+		.then(r => r.json())
+		.catch(() => ({
```
andrewiggins, line 30: Should we let this error bubble so we get telemetry if it starts failing and rely on the client to fallback to default values?

src/lambda/release.js
```diff
@@ -0,0 +1,77 @@
...
+
+function checkStatus(r) {
+	if (!r.ok) {
+		throw new Error(`${r.status}: Request failed for '${r.url}'`);
+	}
+
+	return r;
+}
+
+function memoize(fn) {
```
andrewiggins, line 24: Not sure how long functions persist and it might be overkill for our tiny function, but should we consider a cache that expires entries? I think https://npm.im/men has this capability

JoviDeCroock: That link brings me to an express framework 😅

andrewiggins: lol whoops! I meant https://npm.im/mem

src/lambda/release.js
```diff
@@ -0,0 +1,77 @@
+import fetch from 'node-fetch';
+
+exports.handler = async event => {
+	const { version, url } = await fetchRelease(
+		event.queryStringParameters.repo || 'preactjs/preact'
```
developit, line 5: Nit: can we guard against folks hitting this with arbitrary repo names? Could just allow only `preactjs/` org repos:

```suggestion
		`preactjs/${event.queryStringParameters.repo || 'preact'}`
```

src/lambda/release.js
developit: looks like they basically say stuffing in global scope/context can't be relied on cross-request (timeout is 10s):

https://community.netlify.com/t/how-to-pass-values-between-functions/1292

developit: Idea: can we set caching headers in our existing netlify headers yaml file? they're the ones calling the function, if we tell them to store it for a period of time I think that prevents the function from being executed again.

src/lambda/repo.js
developit: I like that idea. IIRC the client code calling the GitHub API already has handling for non-ok status

src/lambda/repo.js
```diff
@@ -0,0 +1,34 @@
+import fetch from 'node-fetch';
+
+exports.handler = async function(event, context, callback) {
+	const result = await repoInfo(
+		event.queryStringParameters.repo || 'preactjs/preact'
```
developit, line 5: same as [other suggestion](https://github.com/preactjs/preact-www/pull/629/files#diff-78d154ba8815cff0a3e10a5ca1f726bdR5).

src/lib/github.js
```diff
@@ -1,71 +1,28 @@
...
+const baseUrl =
+	process.env.NODE_ENV === 'production'
+		? '/.netlify/functions/'
+		: 'http://localhost:9000/.netlify/functions/';
 
-// TODO: use GraphQL to avoid fetching so much data
-export const repoInfo = memoize(repo =>
-	fetch(`https://api.github.com/repos/${repo}`)
+export const repoInfo = repo =>
+	fetch(`${baseUrl}repo?repo=${repo}`, { mode: 'no-cors' })
```
developit, line 9: `mode:'no-cors'` prevents cors, I think we just want to avoid any cookies:

```suggestion
	fetch(`${baseUrl}repo?repo=${repo}`, { credentials: 'omit' })
```

src/lib/github.js
```diff
@@ -1,71 +1,28 @@
...
-			});
-
-			return {
-				version: releases.length ? releases[0].tag_name : 'unknown',
-				url: releases.length ? releases[0].html_url : '#'
-			};
-		})
+		.then(d => ({
+			version: d.version || 'unknown',
+			url: d.url || '#'
```
developit, line 23: Can probably just go to the repo url here

```suggestion
			url: d.url || 'https://github.com/preactjs/preact'
```

src/lambda/repo.js
JoviDeCroock: We can't do this here since we allow the user to enter anything here

netlify.toml
```diff
@@ -0,0 +1,6 @@
+[build]
+  functions = "dist/lambda"
+[[headers]]
```
JoviDeCroock, line 3: @developit

src/lambda/release.js
```diff
@@ -0,0 +1,76 @@
...
+
+	return {
+		statusCode: 200,
+		body: JSON.stringify({
+			version,
+			url
+		}),
+		headers: {
+			'Cache-Control': 'public, s-maxage=1800',
+			'Access-Control-Allow-Origin': 'http://localhost:8080'
```
JoviDeCroock, line 16: This is to make stuff work in `localhost` a better alternative would be to use `netlify-dev` but that requires people to log into netlify which doesn't feel ideal for an OSS project

marvinhagemeister: Agree, that's a better compromise 👍

src/lambda/repo.js
developit: Whoops good point haha

PR #643, opened 2020-08-08
Remove color theme switcher

marvinhagemeister:
All major browsers have great support for prefers-color-scheme by now. Based on that we'll apply the theme the user marked as preferred in their browser. In other words: We don't need an manual way to do that anymore :tada:


developit:
I'll miss the theme switcher for screenshots but I'm definitely alone there, haha. Nice size savings for CSS too.

PR #654, opened 2020-08-19
Add testing page for @testing-library/preact

marvinhagemeister:
This PR adds a new page for `@testing-library/preact` to the docs.

Probably needs a second read through to get rid of typos/weird sounding sentences. Wrote it down in a single sitting late in the evening 😅

### DIFF
--- a/content/en/guide/v10/preact-testing-library.md
+++ b/content/en/guide/v10/preact-testing-library.md
@@ -5,11 +5,11 @@ description: 'Testing Preact applications made easy with testing-library'
 
 # Testing with Preact Testing Library
 
-The [Preact Testing Library](https://github.com/testing-library/preact-testing-library) is a lightweight wrapper around `preact/test-utils` to verify the rendered DOM. This helps in avoiding including internal implementation details in your tests, thus making them easier to maintain and more resilient against refactorings of your components.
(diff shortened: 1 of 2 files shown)


content/en/guide/v10/preact-testing-library.md
```diff
@@ -0,0 +1,209 @@
+---
+name: Testing with Preact Testing Library
+description: 'Testing Preact applications made easy with testing-library'
+---
+
+# Testing with Preact Testing Library
+
+The [Preact Testing Library](https://github.com/testing-library/preact-testing-library) is a lightweight wrapper around `preact/test-utils` to verify the rendered DOM. This helps in avoiding including internal implementation details in your tests, thus making them easier to maintain and more resilient against refactorings of your components.
+
```
38elements, line 9: > This helps in avoiding including internal implementation details in your tests, thus making them easier to maintain and more resilient against refactorings of your components.

Would you tell me that the above sentence mean the following sentence?

This reduces the impact which the component implementation effects testing. Therefore, it requires fewer changes for test code when you refactor the component.

marvinhagemeister: Close, the main idea is that you just check the DOM after a render if it matches the expected output. So your tests are not "abstract" at all and more closely match what would actually happen if the component would be used.

developit:
This is fantastic.

PR #657, opened 2020-08-27
Fix bug in v10 upgrade guide example.

mikaturk:
Moved the ++ to the front of counter variables.


marvinhagemeister:
Agree with @developit , `+ 1` is probably the notation that's the most beginner friendly.

mikaturk:
> Agree with @developit , `+ 1` is probably the notation that's the most beginner friendly.

I agree too, ++ still affects the original variable so +1 is better.

marvinhagemeister (approved):
Sweet, thank you so much for your PR 🙌

PR #660, opened 2020-08-29
Fix intro text for Preact Testing Library guide

pastelmind:
I replaced some sentences with grammar errors, and added a better explanation of what Preact Testing Library does.


marvinhagemeister (approved):
Awesome, this reads much more fluently! Thank you so much for your PR :+1:

PR #667, opened 2020-08-30
cursor change from text select to default

darvishzadeh:
main title seems to have a text selectable cursor, yet nothing can be selected.
this PR removes pointer events to avoid confusion


developit:
Thanks!

PR #670, opened 2020-08-31
Copy changes

darvishzadeh:
A full sentence receives an ending dot.
A title or statement does not get an ending dot.
An action request can get an exclamation mark.

You can also see this as a proposal.


marvinhagemeister (approved):
Sounds good to me! Thank you so much for your PR 🙌

PR #673, opened 2020-09-04
Renderer Options -> Option Hooks

38elements:
There are `Options`, `Option Hooks` and `Renderer Options`. 
IMHO, I think it is better to be unified into `Option Hooks`.


marvinhagemeister (approved):
Thanks for the PR 🎉

PR #679, opened 2020-09-10
Fix typo in web-components.md

38elements:
I think `complementary` is correct, not `complimentary`.


marvinhagemeister (approved):
Good catch! Had to look it up in a dictionary to spot the different meanings 👍

PR #681, opened 2020-09-17
Improve content/en/guide/v10/web-components.md

38elements:


content/en/guide/v10/web-components.md
```diff
@@ -72,17 +72,20 @@ function Foo() {
 
 ### Triggering custom events
```
38elements, line 73: @developit 
I don't understand the intent of this section.
Does this section only describe custom events?

content/en/guide/v10/web-components.md
```diff
@@ -72,17 +72,20 @@ function Foo() {
...
+Preact normalizes standard DOM events to lowercase. For example, `onChange` prop and `onchange` prop  are regarded as the same "change" event listener. On the other hand, Custom Events are case-sensitive. For example, `onTabChange` prop and `ontabchange` prop are regarded as different event listeners. `onTabChange` prop is regarded as  a "TabChange" event listener. `ontabchange` prop is regarded as  a "tabchange" event listener.
 
 ```jsx
-// Built-in DOM event: listens for a "click" event
+// Standard DOM event: listens for "click" event
 <input onClick={() => console.log('click')} />
 
-// Custom Element: listens for "TabChange" event (case-sensitive!)
```
38elements, line 81: I think `Custom event` is correct, not `Custom Element`.

content/en/guide/v10/web-components.md
```diff
@@ -72,17 +72,20 @@ function Foo() {
...
-<tab-bar onTabChange={() => console.log('tab change')} />
+// Standard DOM event: listens for "click" event
+<input onclick={() => console.log('click')} />
 
-// Corrected: listens for "tabchange" event (lower-case)
```
38elements, line 84: I think `Custom event` is correct, not `Corrected`.

content/en/guide/v10/web-components.md
```diff
@@ -72,17 +72,20 @@ function Foo() {
...
+Preact normalizes standard DOM events to lowercase. For example, `onChange` prop and `onchange` prop  are regarded as the same "change" event listener. On the other hand, Custom Events are case-sensitive. For example, `onTabChange` prop and `ontabchange` prop are regarded as different event listeners. `onTabChange` prop is regarded as  a "TabChange" event listener. `ontabchange` prop is regarded as  a "tabchange" event listener.
 
 ```jsx
-// Built-in DOM event: listens for a "click" event
+// Standard DOM event: listens for "click" event
 <input onClick={() => console.log('click')} />
 
-// Custom Element: listens for "TabChange" event (case-sensitive!)
-<tab-bar onTabChange={() => console.log('tab change')} />
```
38elements, line 82: I think it is unnecessary to use custom element.

content/en/guide/v10/web-components.md
marvinhagemeister: I can see your point about generalising it to CustomEvents, although I'd keep referring to them as `built-in DOM Events` vs `standard DOM events`. To me the former sentence is more clear. The wording in this PR "normalizes standard DOM events to lowercase" is to me not as on point as the previous sentence as we strip the `on` prefix, convert the remainder to lowercase and do a `prop in dom` check to see if it is a built-in event.

content/en/guide/v10/web-components.md
marvinhagemeister: In theory yes, in practice Custom Events are only used with Custom Elements together, so it makes sense for us to include the most commonly used way in our docs.

content/en/guide/v10/web-components.md
marvinhagemeister: I think the point of the example here was that users may use `onTabChange` out of habit and be confused why it's not working. So in a way the lowercase variant "fixes" that.

marvinhagemeister (commented):
Left some comments, let me know what you think 👍

38elements:
IMHO, In this section, I think it is better to focus on the difference in how `Standard DOM Event` and `Custom event` are treated by Preact.

PR #689, opened 2020-10-01
Fix 'Run in REPL' not fixed while scrolling

### LINKED ISSUE #530: `Run in REPL` button doesn't stick to the top-right position in code-blocks(on Docs)
It just scrolls as we scroll horizontally

blenderskool:
## Original Issue

In the code blocks, the `Run in Repl` button did not stay fixed in top right corner while horizontally scrolling. This is especially more noticeable on mobile devices.

### Changes in this fix
A container `<div>` was added to the code block with some style refactoring to keep the `Run in REPL` button fixed at the top right corner even while horizontally scrolling. From my tests, it is not breaking other code blocks (I may be wrong though)

Fixes #530

### DIFF
--- a/src/components/code-block/index.js
+++ b/src/components/code-block/index.js
@@ -73,14 +73,16 @@ function HighlightedCodeBlock({ code, lang, ...props }) {
 	const htmlObj = useMemo(() => ({ __html: html }), [html]);
 
 	return (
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Awesome, thank you so much for the PR🙌

PR #691, opened 2020-10-02
Add padding top to active link targets (headlines)

marzepani:
When navigating to page sections, the selected headline is hidden by the header bar.
This PR adds a `:target` selector with some extra padding on top to keep them visible.

Example: **useState** https://preactjs.com/guide/v10/hooks/#usestate

without `:target` padding

with `:target` padding

### DIFF
--- a/src/style/markdown.less
+++ b/src/style/markdown.less
@@ -373,6 +373,15 @@
 		vertical-align: super;
 		line-height: 0;
 	}
+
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
TIL, about using the `:target` selector for this. This is fantastic. thank you so much for the PR 🙌

PR #692, opened 2020-10-03
Fix content/en/guide/v10/api-reference.md

38elements:
Since `children` is not always an array.


marvinhagemeister (approved):
Good catch 👍

PR #701, opened 2020-11-02
Improve content/en/guide/v10/preact-testing-library.md

38elements:
Since it is difficult for me understand the meaning of this sentence,
I changed it to make it easier to understand for me.

### DIFF
--- a/content/en/guide/v10/preact-testing-library.md
+++ b/content/en/guide/v10/preact-testing-library.md
@@ -128,34 +128,6 @@ test('should be able to sign in', async () => {
 })
 ```
 
-Sometimes using text content directly creates friction when the content changes a lot, or if you use an internationalization framework that translates text into different languages. You can work around this by treating text as data that you snapshot, making it easy to update but keeping the source of truth outside the test.
-
-```jsx
-test('should be able to sign in', async () => {
-  render(<MyLoginForm />);
-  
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/preact-testing-library.md
```diff
@@ -140,15 +140,15 @@ test('should be able to sign in', async () => {
...
-Even if you don't use a translation framework, you can keep your strings in a separate file and use the same strategy as in the example below:
+Supposing `translate()` which uses a key and its corresponding string in a separate file for internationalization, it can be tested as below:
 
 ```jsx
 test('should be able to sign in', async () => {
   render(<MyLoginForm />);
 
-  // We can use our translation function directly in the test
   const label = translate('signinpage.label', 'en-US');
-  // Snapshot the result so we know what's going on
```
38elements, line 151: The description of the snapshot is not explicit.

marvinhagemeister:
I wonder if we can make the docs agnostic to the test runner. It seems to me like the `toMatchInlineSnapshot` is not really necessary here.

38elements:
@marvinhagemeister 
Thank you for reply. 
`toMatchInlineSnapshot` is the function of Jest.
I think it is unnecessary to explain in this document.
Therefore, I removed this description.

PR #705, opened 2020-11-06
Add content/zh/guide/v10/getting-started.md

Nz-JiangHui:


rschristian:
Apologies, this was missed.

I'm going to close it out though, as it looks like it did eventually land in another PR. If you have any suggestions for alterations, you can make them here: https://github.com/preactjs/preact-www/blob/master/content/zh/guide/v10/getting-started.md

PR #706, opened 2020-11-06
Aliasing react to preact/compat using only NPM

MrMaxie:
Well, I found simplest way to create alias between `react` and `preact/compat`, I tested it already with stack:
- of course Preact
- TypeScript
- MobX
- MobX-React
- ESBuild (this builder cannot create own aliases so only just follow requires like plain JS)

And I think it will be helpful for people to know this way


developit:
Hiya

We actually provide npm packages for this so there's no need for the extra files:

```
{
  "dependencies": {
    "react": "npm:@preact/compat",
    "react-dom": "npm:@preact/compat"
  }
}
```

PR #708, opened 2020-11-08
Fix typo in content/en/guide/v10/preact-testing-library.md

38elements:


content/en/guide/v10/preact-testing-library.md
```diff
@@ -201,7 +201,7 @@ import { FooContext } from './foo';
 const history = createMemoryHistory();
 
 export function render(vnode) {
-  return render(
```
38elements, line 204: > we recommend creating a custom `render` function by wrapping the one from `@testing-library/preact`.

`originalRender` function should be wrapped by `render` function.

Issue #713, opened 2020-11-15
Search for "EventListener" yields mixed results to v8 and v10


simon04:
![2020-11-15-184848](https://user-images.githubusercontent.com/782446/99192591-e7e1f680-2773-11eb-9f17-d39b72812ffe.png)

1. links to https://preactjs.com/guide/v8/differences-to-react/#whats-missing
2. links to https://preactjs.com/guide/v10/differences-to-react/#main-differences

The page I wanted to find, https://preactjs.com/guide/v10/hooks#useeffect, is missing in the search results, despite having `window.addEventListener` and `window.removeEventListener` in a code section.

developit:
Ah good point - I'm not sure if there's a way to tell Algolia to bias towards V10 results, will have to check!

developit:
It looks like this should be fixed now! Please re-open if it's still an issue.

simon04:
Thank you for your prompt reaction and for resolving this issue! :-)

PR #715, opened 2020-11-16
docs(hooks): describe event listener

simon04:
Intro for `useEffect`/`addEventListener` example. Mention "event listener" in the text (not only in the code) for algolia search (see #713).


rschristian:
Sorry this went unnoticed.

I think this is fine as-is, tbh. It's meant to be an example of `useEffect`, not necessarily event listeners.

PR #717, opened 2020-11-25
Fixes TypeScript documentation.

MicahZoltu:
There may be a more preferred way to do with the change in behavior in TSC, but I believe this is a reasonable iterative improvement as at least now the instructions will actually work, even if they perhaps aren't ideally laid out.


marvinhagemeister (approved):
Good catch! Thank you so much for the PR 🎉

PR #703, opened 2020-11-04
Add TypeScript documentation, fixes #503

ddprrt:
Hi all 👋

This is the a draft addressing #503

The content is mostly adapted from my [TypeScript + React guide](https://fettblog.eu/typescript-react/) I published two years ago, without the fluff and rephrased to suit proper documentation. It can be seen as an accompanying guide to the regular documentation, focusing on how to properly type so the compiler doesn't complain (which can be ... a lot!).

Things done:
- Documenting types for functional and class components
- Proper types for refs and events
- A section on hooks
- Proper typings for context

Things to do:
- A section with exposed types. I currently browse through index.d.ts to get a feeling for all types used and work on scenarios on how to use them in consumer's code.
- Re-check examples if there is a more up to date version. I also want to prevent people and SE crawlers to mistake this as a duplication of my original guide
- If ok, I want to add a section with further reading to Component type design patterns (e.g. https://fettblog.eu/typescript-react-component-patterns/)

Note: For proper syntax highlighting I had to add `tsx` and `typescript` to Prism.js. I used the link in the prism.js comment to get the original configuration, and added two more languages (the new link is again in prism.js)

PR #718, opened 2020-11-26
Fix content/en/guide/v10/typescript.md

38elements:
This pull request is for reviewing #716.  
Japanese version is added this change to.
I think #703  includes many repeating, incorrect, unnecessary things.


ddprrt:
I think it's not only a type for children that's added to `props` but also the property `children` in general. Worth mentioning?

38elements:
To know what a type for `children` is added to, I think it is necessary.

38elements:
> I think it's not only a type for children that's added to props but also the property children in general. Worth mentioning?

```
`FunctionComponent` also adds a type for `children`:
```

Does the original description make sense?

content/en/guide/v10/typescript.md
```diff
@@ -7,8 +7,6 @@ description: "Preact has built-in TypeScript support. Learn how to make use of i
 
 Preact ships TypeScript type definitions, which are used by the library itself! 
 
-When you use Preact in a TypeScript-aware editor (like VSCode), you can benefit from the added type information while writing regular JavaScript. If you want to add type information to your own applications, you can use [JSDoc annotations](https://fettblog.eu/typescript-jsdoc-superpowers/), or write TypeScript and transpile to regular JavaScript. This section will focus on the latter.
```
38elements, line 10: This is a description of the VSCode features and methods of adding types.
I think it is unnecessary in Preact document.
Therefore, I removed this line.

ddprrt:
You're right, this might be misleading in the original version as well. What about:

```
Preact ships a `FunctionComponent` type to annotate anonymous functions. `FunctionComponent` adds a property `children` to the type for `props`:
```

This should be correct.

content/en/guide/v10/typescript.md
```diff
@@ -92,7 +90,7 @@ function Greeting({ name = "User" }: GreetingProps) {
 }
 ```
 
-Preact also ships a `FunctionComponent` type to annotate anonymous functions. `FunctionComponent` also adds a type for `children`:
+Preact ships a `FunctionComponent` type to annotate anonymous functions. `FunctionComponent` adds a type for `children` to `props`:
```
38elements, line 93: This describes what a type for `children` is added to.
https://github.com/preactjs/preact/blob/f955cfcceb50d504faa7c05a95065512cf900e57/src/index.d.ts#L72-L90

content/en/guide/v10/typescript.md
```diff
@@ -170,7 +170,7 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
 
 ## Typing events
 
-Preact emits regular DOM events. As long as your TypeScript project includes the `dom` library (set it in `tsconfig.json`), you have access to all event types that are available in your current configuration.
```
38elements, line 180: > In Preact core, `onChange` is the standard [DOM change event]

There is "standard DOM event" in other page.
Therefore, I change to "standard DOM event".
Element emits event.

content/en/guide/v10/typescript.md
```diff
@@ -170,7 +170,7 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
 
 ## Typing events
 
-Preact handles standard DOM events. As long as your TypeScript project includes the `dom` library (set it in `tsconfig.json`), you have access to all event types that are available in your current configuration.
```
38elements: `lib` includes `DOM`.
`standard DOM events types` is more explicit.

content/en/guide/v10/typescript.md
```diff
@@ -187,7 +187,7 @@ export class Button extends Component {
 }
 ```
 
-You can restrict event handlers by adding a type annotation for `this` to the function signature as the first argument. This argument will be erased after transpilation.
```
38elements, line 197: This content is an explanation about [this parameters](https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters).
Therefore, This is not appropriate in Preact document.

content/en/guide/v10/typescript.md
```diff
@@ -144,47 +141,43 @@ type ExpandableState = {
...
 
-Preact emits regular DOM events. As long as your TypeScript project includes the `dom` library (set it in `tsconfig.json`), you have access to all event types that are available in your current configuration.
+Preact handles standard DOM events. When [lib](https://www.typescriptlang.org/tsconfig#lib) of TypeScript compiler options includes `DOM`, you can use standard DOM events types.
 
 ```tsx
 export class Button extends Component {
   handleClick(event: MouseEvent) {
     event.preventDefault();
     if (event.target instanceof HTMLElement) {
-      alert(event.target.tagName); // Alerts BUTTON
```
38elements, line 187: Another one uses `console.log()`.
It is inconsistent.

content/en/guide/v10/typescript.md
```diff
@@ -144,47 +141,43 @@ type ExpandableState = {
...
     this.state = {
       toggled: false
     };
   }
-  // `this.props.title` is string due to ExpandableProps
   render() {
     return (
       <div class="expandable">
         <h2>
+          // `this.props.title` is string type by ExpandableProps.
```
developit, line 157: I need to check if our syntax highlighter can handle line comments in JSX

content/en/guide/v10/typescript.md
```diff
@@ -144,47 +141,43 @@ type ExpandableState = {
...
   }
 }
 ```
 
-Class components include children by default, typed as `ComponentChildren`.
-
 ## Typing events
 
-Preact emits regular DOM events. As long as your TypeScript project includes the `dom` library (set it in `tsconfig.json`), you have access to all event types that are available in your current configuration.
+Preact handles standard DOM events. When [lib](https://www.typescriptlang.org/tsconfig#lib) of TypeScript compiler options includes `DOM`, you can use standard DOM events types.
```
developit, line 173:
```suggestion
Preact uses standard DOM events. Include `"DOM"` in TypeScript's [lib](https://www.typescriptlang.org/tsconfig#lib) compiler option to enable standard DOM event types.
```

content/en/guide/v10/typescript.md
developit: Good call!

content/en/guide/v10/typescript.md
```diff
@@ -194,7 +187,7 @@ export class Button extends Component {
 }
 ```
 
-You can restrict event handlers by adding a type annotation for `this` to the function signature as the first argument. This argument will be erased after transpilation.
+You can specify a type of `EventTarget` by using [this parameters](https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters), as below.
```
developit, line 190:
```suggestion
You can provide a more specific type than `EventTarget` by using [this parameters](https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters):
```

content/en/guide/v10/typescript.md
```diff
@@ -130,7 +127,7 @@ function Card({ title, children }: ChildrenProps) {
 
 ### Class components
 
-Preact's `Component` class is typed as a generic with two generic type variables: Props and State. Both types default to the empty object, and you can specify them according to your needs.
+Preact's `Component` class is a generic type. This type has two generic type variables which correspond Props and State. These generic type variables are empty object in default. You can define type and pass one as generic type variable, as below.
```
developit, line 130:
```suggestion
Preact's `Component` class is a generic type. This type has two generic type variables which correspond Props and State. These generic type variables are empty objects by default, and can be defined as follows:
```

content/en/guide/v10/typescript.md
developit: A thought: may we could have this just be a short "tip" like this:

Screen Shot 2020-12-07 at 10 31 46 AM

the markup I used there was:

```md
> 💁 You can also use Preact's type definitions in plain JavaScript with [JSDoc annotations](https://fettblog.eu/typescript-jsdoc-superpowers/).
```

content/en/guide/v10/typescript.md
```diff
@@ -194,16 +189,14 @@ export class Button extends Component {
...
 export class Button extends Component {
-  // Adding the this argument restricts binding
+  // Define `this parameters`
   handleClick(this: HTMLButtonElement, event: MouseEvent) {
     event.preventDefault();
-    if (event.target instanceof HTMLElement) {
-      console.log(event.target.localName); // "button"
-    }
+    console.log(this.tagName); // "BUTTON"
   }
```
38elements, line 200: I think this example should use `this`.
Since `this` is `HTMLButtonElement`, `instanceof` is unnecessary.
Another example uses `tagName`.

ddprrt: I agree, this is much more concise and showcases this binding much better 👍

content/en/guide/v10/typescript.md
```diff
@@ -216,19 +210,15 @@ export class Button extends Component {
...
 
 ```tsx
 import { h, Component, createRef } from "preact";
 
 class Foo extends Component {
+  // `this.ref.current` is HTMLAnchorElement type.
   ref = createRef<HTMLAnchorElement>();
 
-  componentDidMount() {
-    // current is of type HTMLAnchorElement
```
38elements, line 228: `this.ref.current` is `div`.
This comment is confusing.

ddprrt: Good catch!

content/en/guide/v10/typescript.md
```diff
@@ -216,29 +210,23 @@ export class Button extends Component {
...
+    //       💥 Error! `this.ref` can only be passed to Anchor Element's `ref` attribute.
   }
 }
 ```
 
-This helps a lot if you want to make sure that the elements you `ref` to are input elements that can be e.g. focussed.
-
 ## Typing context
 
 `createContext` tries to infer as much as possible from the intial values you pass to:
```
38elements, line 232: TypeScript infers, not `createContext`.

content/en/guide/v10/typescript.md
```diff
@@ -56,13 +57,13 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
 
 Rename your `.jsx` files to `.tsx` for TypeScript to correctly parse your JSX.
 
-## Typing components
+## Typing Components
 
-There are different ways to type components in Preact. Class components have generic type variables to ensure type safety. TypeScript sees a function as functional component as long as it returns JSX. There are multiple solutions to define props for functional components.
+There are different ways to type components in Preact. Class components have generic type parameters to ensure type safety. TypeScript sees a function as functional component as long as it returns JSX. There are multiple solutions to define props for functional components.
 
-### Function components
```
38elements, line 63: They are "Functional Components" and "Class Components" in other page.
https://github.com/preactjs/preact-www/blob/922098858eba2769900cc909b18b7b27e5487a49/content/en/guide/v10/components.md#functional-components

content/en/guide/v10/typescript.md
```diff
@@ -92,7 +93,7 @@ function Greeting({ name = "User" }: GreetingProps) {
 }
 ```
 
-Preact also ships a `FunctionComponent` type to annotate anonymous functions. `FunctionComponent` also adds a type for `children`:
```
38elements, line 95: `FunctionComponent` type corresponds all functional components, not only an anonymous function.
This sentence is confusing.

content/en/guide/v10/typescript.md
```diff
@@ -56,13 +57,13 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
 
 Rename your `.jsx` files to `.tsx` for TypeScript to correctly parse your JSX.
 
-## Typing components
+## Typing Components
 
-There are different ways to type components in Preact. Class components have generic type variables to ensure type safety. TypeScript sees a function as functional component as long as it returns JSX. There are multiple solutions to define props for functional components.
```
38elements, line 61: > TypeScript sees a function as functional component as long as it returns JSX.

Unless a function is matched `FunctionComponent` or `FunctionalComponent`, it is not considered a functional components.
This is incorrect.

38elements: I think this line is unnecessary.

38elements: Since there is the same content at below, I think this line is unnecessary.

content/en/guide/v10/typescript.md
```diff
@@ -79,20 +78,7 @@ function MyComponent({ name, age }: MyComponentProps) {
...
-```tsx
-type GreetingProps = {
-  name?: string; // name is optional!
-}
-
-function Greeting({ name = "User" }: GreetingProps) {
-  // name is at least "User"
-  return <div>Hello {name}!</div>
-}
-```
```
38elements, line 93: This is explanation for Default parameters.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
Therefore, This is not appropriate in TypeScript page of Preact document.

content/en/guide/v10/typescript.md
```diff
@@ -56,59 +57,57 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
...
 
-function MyComponent({ name, age }: MyComponentProps) {
+function MyComponent({ name, age }: Props) {
   return (
     <div>
-      My name is {name}, I am {age.toString()} years old.
+      My name is {name}, I am {age} years old.
     </div>
   );
 }
```
38elements, line 90: This example should be similar to the example of anonymous function.

content/en/guide/v10/typescript.md
```diff
@@ -56,69 +57,53 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
...
-}
+`children` is `ComponentChildren` type. `children` as `ComponentChildren` type is added to props in `RenderableProps` type:
 
-function Greeting({ name = "User" }: GreetingProps) {
-  // name is at least "User"
-  return <div>Hello {name}!</div>
-}
+```ts
+type RenderableProps<P, RefType = any> = P & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<RefType> }>;
 ```
```
38elements, line 97: Most React library use  `React.FC`.
In Preact, most library use `FunctionComponent`.
Using `FunctionComponent` method is general.
Therefore, This example is unnecessary.

content/en/guide/v10/typescript.md
```diff
@@ -81,7 +81,7 @@ type Props = {
   age: number;
 };
 
-function MyComponent({ name, age }: Props) {
```
38elements: Most React library use  `React.FC`.
In Preact, most library use `FunctionComponent`.
Using `FunctionComponent` method is general.
Therefore, This example is unnecessary.

content/en/guide/v10/typescript.md
```diff
@@ -144,47 +129,43 @@ type ExpandableState = {
...
         <div hidden={this.state.toggled}>{this.props.children}</div>
       </div>
     );
   }
 }
 ```
 
-Class components include children by default, typed as `ComponentChildren`.
-
-## Typing events
```
38elements, line 178: I think it is necessary to describe about `TargetedEvent`, since `TargetedEvent` does not exist in React.

content/en/guide/v10/typescript.md
```diff
@@ -56,69 +57,53 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
...
-`children` is of type `ComponentChildren`. You can specify children on your own using this type:
-
-
-```tsx
-import { h, ComponentChildren } from "preact";
-
-type ChildrenProps = {
+type Props = {
   title: string;
-  children: ComponentChildren;
```
38elements, line 118: Since `children` as `ComponentChildren` type is added to props in `RenderableProps` type, this line is unnecessary.

content/en/guide/v10/typescript.md
```diff
@@ -178,22 +253,38 @@ export class Button extends Component {
 You can provide a more specific type than `EventTarget` by using [this parameters](https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters):
 
 ```tsx
+import { h, Component } from 'preact';
+
 export class Button extends Component {
   // Define `this parameters`
   handleClick(this: HTMLButtonElement, event: MouseEvent) {
-    event.preventDefault();
```
38elements, line 203: This line is unnecessary. Therefore, I removed it.

content/en/guide/v10/typescript.md
```diff
@@ -140,30 +214,31 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
...
+## Events
 
 Preact uses standard DOM events. Include `"DOM"` in TypeScript's [lib](https://www.typescriptlang.org/tsconfig#lib) compiler option to enable standard DOM event types.
 
 ```tsx
+import { h, Component } from 'preact';
+
 export class Button extends Component {
   handleClick(event: MouseEvent) {
-    event.preventDefault();
```
38elements, line 185: This line is unnecessary. Therefore, I removed it.

content/en/guide/v10/typescript.md
```diff
@@ -194,52 +250,60 @@ export class Button extends Component {
...
-    if (event.target instanceof HTMLElement) {
-      console.log(event.target.localName); // "button"
-    }
+    console.log(this.tagName); // "BUTTON"
   }
 
   render() {
-    return (
-      <button onClick={this.handleClick}>{this.props.children}</button>
-    );
```
38elements, line 212: This `(` and `)` are unnecessary. Therefore, I removed it.

content/en/guide/v10/typescript.md
```diff
@@ -194,52 +251,59 @@ export class Button extends Component {
...
+  render() {
+    return <button onClick={this.handleClick}>{this.props.children}</button>;
   }
 }
 ```
 
-## Typing references
+## References
 
-The `createRef` function is also generic, and lets you bind references to element types. In this example, we ensure that the reference can only be bound to `HTMLAnchorElement`. Using `ref` with any other element lets TypeScript thrown an error:
```
38elements, line 219: >  Using `ref` with any other element lets TypeScript thrown an error.

This is explanation about TypeScript `TypeError`.
I think it is unnecessary in Preact document.
Therefore, I removed this sentence.

content/en/guide/v10/typescript.md
```diff
@@ -194,52 +251,59 @@ export class Button extends Component {
...
 
-  componentDidMount() {
-    // current is of type HTMLAnchorElement
-    console.log(this.ref.current);
-  }
-
   render() {
     return <div ref={this.ref}>Foo</div>;
     //          ~~~
-    //       💥 Error! Ref only can be used for HTMLAnchorElement
```
38elements, line 235: This content is repeating.

content/en/guide/v10/typescript.md
```diff
@@ -173,18 +232,17 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
 }
 ```
 
-Class components include children by default, typed as `ComponentChildren`.
```
38elements, line 176: `children?` is default, not `children`.

```ts
type RenderableProps<P, RefType = any> = P & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<RefType> }>;
```

38elements:
> I think it's not only a type for children that's added to props but also the property children in general. Worth mentioning?

I sent this pull request because I thought it was worth it.
Is a person who wrote this existing  description  a person who deserves to say "Is this worth it?" to others?

content/en/guide/v10/typescript.md
```diff
@@ -100,7 +100,7 @@ const MyComponent: FunctionComponent<Props> = function ({ name, age }) {
 type RenderableProps<P, RefType = any> = P & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<RefType> }>;
 ```
 
-When `children` is required, it need to be specified:
+Since `children` in `RenderableProps` type is `children?: ComponentChildren`, `children` need to be specified when `children` is required:
```
38elements, line 103: This clarifies the reason for specifying `children`.

content/en/guide/v10/typescript.md
```diff
@@ -258,72 +331,7 @@ const AppContext = createContext({
...
-        lang: "de",
-        authenticated: true
-      }}
-    >
-    {}
-      <ComponentThatUsesAppContext />
-    </AppContext.Provider>
-  );
-}
-```
```
38elements, line 280: This is explanation about TypeScript `TypeError`.
I think it is unnecessary in Preact document.
Therefore, I removed this sentence.

content/en/guide/v10/typescript.md
```diff
@@ -258,72 +331,7 @@ const AppContext = createContext({
...
-      value={{
-        lang: "de",
-        ...appContextDefault
-      }}
-    >
-      <ComponentThatUsesAppContext />
-    </AppContext.Provider>
-  );
-}
-```
```
38elements, line 299: This is mere explanation about how to pass values to `props` using destructuring assignment.
I think it is not suitable in this page.
Therefore, I removed this sentence.

content/en/guide/v10/typescript.md
```diff
@@ -258,72 +331,7 @@ const AppContext = createContext({
...
-      value={{
-        lang: "de"
-      }}
-    >
-      <ComponentThatUsesAppContext />
-    </AppContext.Provider>
-  );
-```
-
-All values become optional, so you have to do null checks when using them.
```
38elements, line 324: This is explanation about `Partial` type.
I think it is unnecessary in Preact document.
Therefore, I removed this sentence.

content/en/guide/v10/typescript.md
```diff
@@ -350,7 +358,7 @@ const Counter = ({ initial = 0 }) => {
 
 `useEffect` does extra checks so you only return cleanup functions.
```
38elements, line 359: It seems this line is mere explanation about `useEffect`.  Since Hooks page is suitable, this line is unnecessary.

content/en/guide/v10/typescript.md
```diff
@@ -350,15 +358,13 @@ const Counter = ({ initial = 0 }) => {
...
-```typescript
+```ts
 useEffect(() => {
   const handler = () => {
     document.title = window.innerWidth.toString();
   };
   window.addEventListener("resize", handler);
 
-  // ✅  if you return something from the effect callback
-  // it HAS to be a function without arguments
```
38elements, line 361: It seems this line is mere explanation about `useEffect`. 
No one passes an argument since this function does not accept arguments.
Therefore, this lines is unnecessary.

content/en/guide/v10/typescript.md
marvinhagemeister: This is the formatting that prettier applies by default and we have that on other pages of our docs

content/en/guide/v10/typescript.md
marvinhagemeister: I think the section was great and we should keep it. People switching from prop types over to TS will naturally ask these questions as most linters treat prop-types and default props as the same. We should cater to those users too :+1:

marvinhagemeister (changes requested):
This PR has gotten a bit out of hand and is impossible to merge. Most of the changes here are made under the assumption that there should be a strict divide between what should be in the TS docs, on MDN or in ours. The line is much more blurry in reality and I vastly prefer the previous approach of including some tips in our docs, even if they can be found in detail in others scattered over the internet.

I know this is just a draft, but we should strife to make smaller PRs. When they are more about corrections they become easier to review too :+1:

38elements:
@marvinhagemeister 
This pull request's first title is "Improve content/en/guide/v10/typescript.md".
This pull request was  only one-line change at first.
(This pull request  was squashed.)
https://github.com/preactjs/preact-www/pull/718#discussion_r531498923
I added fellowing message at first.
```
This describes what a type for children is added to.
https://github.com/preactjs/preact/blob/f955cfcceb50d504faa7c05a95065512cf900e57/src/index.d.ts#L72-L90
```

I think this is a looking down way of saying it:

"I think it's not only a type for children that's added to props but also the property children in general. Worth mentioning?"

I can accept the following comments:

"I think it's not only a type for children that's added to props but also the property children in general. Therefore, I think this change is unnecessary."

I have already explained. And I sent it because I think it is worth it. I do not think it is a necessary question.
I think his comments are strange and inconsistent. Only I am commented like this in Preact community.

I think you have not been never commented like this in Preact community.
Have you been commented in here?

content/en/guide/v10/typescript.md
```diff
@@ -56,86 +57,147 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
...
+```
 
-Preact's `Component` class is typed as a generic with two generic type variables: Props and State. Both types default to the empty object, and you can specify them according to your needs.
+You can implement a class component:
 
 ```tsx
+import { h, Component } from 'preact';
+
 // Types for props
+// in the case `children` is required.
```
38elements, line 197: This line is unnecessary. I made mistake.

content/en/guide/v10/typescript.md
```diff
@@ -56,86 +57,147 @@ If you use TypeScript within a Babel toolchain, set `jsx` to `preserve` and let
...
 type ExpandableProps = {
   title: string;
+  children: ComponentChildren;
```
38elements, line 200: This line is unnecessary. I made mistake.

PR #721, opened 2020-12-03
Adds config to use preact/compat with Snowpack

Macavirus:
Snowpack defines its own specific aliasing system which must be configued to use with preact/compat.


marvinhagemeister (approved):
This is great 🙌

PR #723, opened 2020-12-07
refactor: Suggest NPX over global installation

rschristian:
For a few months now the CLI has suggested the use of NPX over global installation as NPX would keep the CLI up-to-date and solve some of the permission issues that popped up (`sudo` and the like). https://github.com/preactjs/preact-cli#usage

This PR removes installation from the instructions to better match current best practices. Of these languages I only speak English and some Spanish so for Portuguese and Japanese I copied from the header meta and checked with Google Translate. I'm hoping that since these are simple translations (just "Getting Started") that they're safe.

### DIFF
--- a/content/en/cli/getting-started.md
+++ b/content/en/cli/getting-started.md
@@ -4,15 +4,7 @@ permalink: '/cli/getting-started'
 description: 'Getting started with preact CLI documentation'
 ---
 
-# Installation
-
-To get started with Preact CLI, install it from [npm](https://npmjs.com/package/preact-cli):
-
-```shell
-npm i -g preact-cli
-```
-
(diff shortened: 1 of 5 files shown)


developit:
Heyo - we have to land english version first so that the translations can be updated by the folks we have for those languages. Are you okay removing the 3 other localized versions?

marvinhagemeister:
Was gonna write something similar. At least when it comes to the German language it's a bit of a hit and miss to use Google Translate for that. Easier if we update translations in a separate PR 👍

rschristian:
Sorry, gave it a shot as it was a 1-2 words copied from the header. Been reverted though.

marvinhagemeister (approved):

PR #724, opened 2020-12-14
Update the current release version in header.

xorgy:
I opened the site in my browser today and got confused when I saw a version number from last year in the header, turns out I had JS turned off.


marvinhagemeister (approved):
Awesome, thank you so much for the PR 🙌

PR #728, opened 2020-12-29
add info for onInput vs onChange

borakuldeep:


rschristian:
Sorry, this slipped through the cracks. I'm going to close this out though as it's documented already in the current version of the docs: https://preactjs.com/guide/v10/differences-to-react#use-oninput-instead-of-onchange

PR #730, opened 2020-12-29
Refresh package-lock

andrewiggins:
Lots of old deps were hanging around in the package-lock file. This PR refreshes them. Hopefully nothing breaks since all these changes should be within semver but probably still good to do a cursory glance over the site


developit:
We were on Node 10 for some reason, I bumped to 12 to fix the usage of `[].flat()`. Should move to GH Actions though.

marvinhagemeister (approved):

PR #736, opened 2021-01-10
Fixed typos in web components example

insekticid:
Fixed typos
```
registerElement -> register
Foo -> TextSection
<slot name=""> -> <span slot="">
```


marvinhagemeister (approved):
Good catch 👍

PR #741, opened 2021-01-14
Update CODE_OF_CONDUCT.md

### LINKED ISSUE #740: Forbid worthless comment part 2
@JoviDeCroock
> Preact also ships a `FunctionComponent` type to annotate anonymous functions. `FunctionComponent` also adds a type for `children
> I think it’s not only a type for children that’s added to props but also the property children in general.

This comment is incorrect. Component class constructor's  props does not have children. Since He does not know it, he made mistaked.
https://github.com/preactjs/preact-www/pull/720
Therefore, this comment is no value.

> Why do you feel this comment is offensive towards you?

Most people send pull request because they think it's worth it.
I think it is unnatural to ask if it is worth it.

>>I have not seen you are commented like that comment in here.
> 
> I'm not quite sure what you mean by this

I feel that the comments made by some European newcomer differ depending on whether European or not.

38elements:
This resolves #740.
This change prevents people from making the same mistakes as I do.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -13,6 +13,7 @@ Examples of behavior that contributes to creating a positive environment include
 * Gracefully accepting constructive criticism
 * Focusing on what is best for the community
 * Showing empathy towards other community members
+* Western values are prioritized over others
 
 Examples of unacceptable behavior by participants include:
 


kristoferbaxter:
Closing this issue, since adding this language goes against the goals of the project.

PR #742, opened 2021-01-15
chore: Update contributors list for footer

rschristian:
This list hasn't been update in about a year and a half or so, since it was first created. I think it's due for an update, as quite a few new names have helped out since then. 

Edit: Though now that I look a bit more closely, there's a couple names from the old list that aren't in the new. Is there another list that should be appended in? Not sure how that should be handled.


developit:
I'm guessing this is because of paging?

rschristian:
Not sure. The landing page (I guess you'd call it) for the repo shows 263 contributors, but the actual contributors link is ~100. I think it'll do that for contributions that aren't linked to an account, but that's still a pretty big difference. 

The names I noticed that were different (though there might be others) are kristopherbaxter and reznord, right at the end of the old list. As they're not on the repo's contributors page, and as they're at the end of the list, I'm guessing they were manually added? Not sure.

developit:
Running this in a JS console should return a deduped list of names for all org repos, but it hits GitHub's rate limit (or at least it did for me after running it twice):

https://gist.github.com/developit/d8c53ec368d56ca113b191dd5a3642b4

rschristian:
So there's a *tad* few more names now....

This PR brought the list from 32 to 565!

developit:
now that I think of it, we should inject these at build time to avoid putting all the names in the client-side JS. I can try to come up with a hacky solution...

rschristian:
That did dawn on me once I saw the number of names. 

Could set up some tiny API too to return a random name. The footer is a pretty safe place to have a flash of content as the name changes.

developit:
@rschristian yeah either that or a JSON file that we fetch lazily seems like the way to go here. We can still either "bake in" one name randomly each time the site gets prerendered, or even just have the initial footer not include a name and it only gets injected when changing the URL.

For the API thing, we currently have two Netlify functions, this could be a third? I think I'm slightly inclined to try the simple JSON-file-loaded-via-fetch approach first since it's free. Something like:

**footer.js:**

```diff
+let contributors;

export function useContributors(deps) {
-	const [value, setValue] = useState(CONTRIBS[new Date().getMonth()]);
+	const [value, setValue] = useState(contributors ? contributors[new Date().getMonth()] : undefined);
+	useEffect(() => {
+		fetch('/assets/contributors.json').then(r => r.json()).then(data => contributors = data);
+	}, []);
	useEffect(() => {
-		setValue(CONTRIBS[(Math.random() * (CONTRIBS.length - 1)) | 0]);
+		if (contributors) {
+			setValue(contributors[(Math.random() * (contributors.length - 1)) | 0]);
+		}
	}, deps);
	return value;
}
```

... then later on:

```diff
				<p style="line-height: 1">
					Built by a bunch of{' '}
					<a
						href="https://github.com/preactjs/preact/graphs/contributors"
						target="_blank"
						rel="noopener noreferrer"
					>
						lovely people
-					</a>{' '}
-					like{' '}
+					</a>
+					{contrib && [
+						' like ',
						<a
							href={'https://github.com/' + contrib}
							target="_blank"
							rel="noopener noreferrer"
						>
							@{contrib}
						</a>
+					]}
					.
				</p>
```

rschristian:
Totally forgot about this thing. Will try to figure out something tomorrow. Whoops!

rschristian:
@developit Should (hopefully) be good now. Went with the static JSON file as you suggested. Will be ~4.5kb of transferred JSON but I'm guessing that should be okay.

As a side note, we're up 92 unique new contributors to PreactJS repos since Jan 14 which is pretty cool! ~15% of the total unique contributors to the org have contributed for the first time in the last 95 or so days.

PR #746, opened 2021-01-22
Add 1tuner to Full Apps list

robinbakker:


developit (approved):
Yes! how was this not on the list already :D

PR #751, opened 2021-02-15
Fixed typo in web-components slot example

jonathanheemstra:
Replace closing `</slot>` tag with `</span>`


marvinhagemeister (approved):
Good catch! Thank you so much for the PR 👍

Issue #752, opened 2021-02-20
Wrong link of getting-started in Japanese translate


takurinton:

JoviDeCroock:
Closing as we've resolved the conversation in #753 this comes down to a discrepancy between GH linking between doc-pages and the published equivalent.

PR #753, opened 2021-02-20
fix: preact-cli document link

takurinton:
https://github.com/preactjs/preact-www/issues/752

### DIFF
--- a/content/ja/cli.md
+++ b/content/ja/cli.md
@@ -34,4 +34,4 @@ Preact CLIは以下の機能をデフォルトで使うことができます。
 - preact-router
 - 必要に応じてロードされる1.5kbのfetchとPromiseのpolyfill
 
-詳しく知りたい場合は[はじめに](/cli/getting-started)を読んでください。
+詳しく知りたい場合は[はじめに](https://github.com/preactjs/preact-www/blob/master/content/ja/cli/getting-started.md)を読んでください。


JoviDeCroock:
Hey, thanks for the PR.

I just checked on the [main website](https://preactjs.com/cli) and it looks like this correctly links to the [getting started page](https://preactjs.com/cli/getting-started) when I click the link.

What issues are you encountering, might be an issue that causes a discrepancy between local and published, glad to help out where I can.

takurinton:
@JoviDeCroock 

Thank you for your reply.  
This was originally defined as a relative path.
If it's not my misunderstanding, I think I'll move [here](https://github.com/preactjs/preact-www/blob/master/content/ja/cli/getting-started.md).

JoviDeCroock:
Both should work, if we look at the deploy-preview and live-site both work great. I just tested it locally and both work so let's keep it consistent across the whole site, in the [other documentation pages](https://raw.githubusercontent.com/preactjs/preact-www/master/content/pt-br/cli.md) we also link with `/cli/getting-started.md`

takurinton:
@JoviDeCroock

sorry.
There may be a problem with my environment.
confirm.
Thank you very much.

JoviDeCroock:
@takurinton No problem! If I can help out with environment problems feel free to poke me

takurinton:
@JoviDeCroock

I would like to ask another question.
Is the link in the bottom line [here](https://github.com/preactjs/preact-www/blob/master/content/ja/cli.md) correct [here](https://github.com/preactjs/preact-www/blob/master/content/ja/cli/getting-started.md)?

JoviDeCroock:
which links in particular, if you check the live site they work for me. What issue are you seeing?

takurinton:
@JoviDeCroock


If you click here, the page does not exist like this.

image

JoviDeCroock:
Oh yeah in GitHub that's normal, the documentation is meant to be explored on the live site 😅 let's look at the difference, in [your deploy preview](https://deploy-preview-753--preactjs.netlify.app/cli) this getting-started link now links us to GitHub which isn't the intention, we'd like to keep the existing behavior where we stay on our documentation site, you can compare by clicking the same link [here](https://preactjs.com/cli)

takurinton:
@JoviDeCroock

Oh!!!
Ok!! I understood!! Sooooooooryyyyyyy🙇‍♂️  
Thank you!!!

JoviDeCroock:
No problem, glad I could help out!

developit:
(fwiw I totally wish there was a way we could have both work!)

Issue #754, opened 2021-02-23
Fix Internationalization heading link to english


x86chi:
In Internationalization document, The Table of Content Link is muted to `-`.

This issue was reported from Japanese document (#646). You can reproduce from any language.

<img width="366" alt="Screen Shot 2021-02-23 at 12 57 16 PM" src="https://user-images.githubusercontent.com/5278201/108800542-c6947000-75d6-11eb-913e-6d5d493e7021.png">

In React doucment, Internationalization [heading is wrapped to english](https://ko.reactjs.org/docs/getting-started.html#try-react)

The document in the link above is maintain with english:

```md
## React Element의 타입 지정하기 {#specifying-the-react-element-type}
```

It seems good to unify the links in English. Because there is no permerlink for each language like `ko.preactjs.org`.

developit:
@x86chi it's possible to permalink: https://preactjs.com/?lang=kr

x86chi:
From the next page, the `lang` query string will be hidden.

Navagate to "Differences to React" steps:

1. https://preactjs.com/?lang=ko
2. https://preactjs.com/guide/v10/differences-to-react

I think it is difficult to manually paste and share the query string.

developit:
@x86chi That's a fair point, yep! I had originally gone with the disappearing querystring because we didn't want to assume shared links should point to the same language - that way someone can share a link to documentation and it will detect.

Out of curiosity - did the docs automatically show in Korean for you when you first visited the site?  We try to detect your language based on `navigator.languages`, but it's imperfect. Maybe that could be improved too.

I just re-read your original description and you're totally right about the heading identifiers - we're using a regular expression to convert non-alphanumeric characters to `-` and it doesn't account for non-english characters at all. We can definitely fix that.

x86chi:
> did the docs automatically show in Korean for you when you first visited the site?

It works fine in modern browsers.

From a permerlink perspective, it seemed more semantic to always use a query string.

> we're using a regular expression to convert non-alphanumeric characters to `-`. We can definitely fix that.

Thanks for answer. Can I contribute about it?

PR #756, opened 2021-03-06
Support non-alphanumeric characters in ToC

x86chi:
```js
// Note: character range in regex is roughly "word characters including accented" (eg: bublé)
const id = text
  // korean, Japanese (hiragana)
  .replace(/[^가-힣\u3041-\u3096]/g, '');
```

If want to keep existing intent, the above code might be more suitable.

But i haven't opinion yet, so I remove the one line that erased non-alphanumeric characters.

Reference issue comment: https://github.com/preactjs/preact-www/issues/754#issuecomment-789785108


src/lib/content.js
```diff
@@ -159,7 +159,6 @@ function generateToc(markdown) {
 		const id = text
 			.toLowerCase()
 			.replace(/[\s-]+/g, '-')
```
developit, line 161:
```suggestion
			.replace(/[\s-!()`'"&,]+/g, '-')
```

developit (approved):
Nice! This is an improvement.

PR #762, opened 2021-04-11
Improve dark theme sponsor logos

andrewiggins:
When the deploy preview finishes, check it out on browsers you have as well as the hover effect, and let me know what you think!


andrewiggins:
I don't have a Mac machine, so someone verifying the deploy preview on dark/light Safari would be very helpful!

marvinhagemeister (approved):

PR #780, opened 2021-06-20
Update forms.md: it's -> its

jaller94:


marvinhagemeister (approved):
Good catch, thank you so much for making a PR 🙌

PR #783, opened 2021-07-04
fix: handling lang attribute in html tag

chijure:
handling attribute lang in tag html


rschristian:
Sorry, what does this fix? Seems to all be working as-is.

chijure:
> Sorry, what does this fix? Seems to all be working as-is.

when you first enter the language is English, and the lang = "en", but when you change the language and reload the lang is still "en"

Screen Shot 2021-07-11 at 18 16 42

src/lib/i18n.js
```diff
@@ -19,6 +19,8 @@ export function useLanguage() {
 		[url]
 	);
 
+	setLang(lang);
```
developit, line 22: I believe this will cause an infinite promise loop, or at least a lot of duplicate renders.

rschristian:
Closing in favor of #1104

Sorry for taking so long to address this, but thanks for taking a crack at it!

PR #784, opened 2021-07-17
fix typo, improve wording

jkulton:
- **Typo:** replaced `invoking` with `invoked`
- **Improvement:** removed "one" after "second" since there was no "one" after the word "first" in the same sentence


marvinhagemeister (approved):
Good catch! Thanks for the PR 👍

PR #786, opened 2021-08-12
Add jsx-runtime aliases to Getting Started

Worble:
Adds jsx-runtime aliases to the relevant configuration files on the Getting Started page.
Rollup and Snowpack are noticeably missing the `preact/test-utils` alias - is this intentional or should they also be added?


marvinhagemeister (approved):
Awesome, thank you so much for the PR 🙌

PR #790, opened 2021-08-27
Fix source maps served with wrong content type header

marvinhagemeister:


rschristian:
Going to close this out as post-Vite migration we're no longer building source maps (though our prerenderer is still adding source map comments, which'll need to be fixed upstream)

PR #791, opened 2021-08-29
Use explicit `.json` and `.module.less` extension in import specifiers

marvinhagemeister:


developit (approved):

marvinhagemeister:
Looks like travis hangs :S

PR #794, opened 2021-09-08
Fix Snowpack config for aliasing

nesk:
Nothing much to add, the current configuration cannot be copied/pasted without manually adding `export default` and it can bring confusion.


marvinhagemeister (approved):
Good catch! Thank you so much for the PR 🙌

PR #797, opened 2021-09-26
Fix minor typo in reference guide

jervisfm:
Then was misspelled as than


rschristian (approved):
Thanks for the catch!

PR #801, opened 2021-10-06
Update acronym to match the more common definition

AshKyd:
I noticed this acronym, while clever, isn't the widely held definition for CLI. Per Wikipedia: https://en.wikipedia.org/wiki/Command-line_interface


marvinhagemeister (approved):
Good catch! Thanks for filing a PR 👍

PR #807, opened 2021-10-29
Fix a bug in preact testing library usage example

heatherKoo07:
Current example `expect(screen.textContent).toMatch("Current value: 6");`
    failed with the following error message
        expect(received).toMatch(expected)
        Matcher error: received value must be a string
        Received has value: undefined
    queries like getByText should be used to work with Async APIs
    like waitFor or findBy


developit (approved):

developit:
LGTM, though @marvinhagemeister knows more than I do about testing-library.

PR #809, opened 2021-12-04
Create a tutorial for learning Preact from scratch

JoviDeCroock:
## Intent

I've recently seen a lot of folks coming in to Preact as first-timers to VDom and our docs are historically not too accessible as we assume prior knowledge. This tutorial aims to combat that once and for all.

This "project" will be a story of iterations rather than something we can get right from the start, we'll need some folks evaluating whether this makes sense to them as all of us at the core-team have a bias and are experienced with virtual dom.

## Chapters

- [x] VDom
- [x] Dom-Events
- [x] Components
- [x] State
- [x] Refs
- [x] Context
- [x] Error-boundaries
- [x] Side-effects

One thing I've been considering is to add a "real-world" example after "side-effects" where we fetch some data and then render a list, this would incorporate the `key` keyword as well. 

I'm also contemplating whether we want to "verify" people's solutions optionally, we could just check the output and see whether it checks out. Some of these like the refs one easily allow for us to do this.

## Todos

- [x] improve on styling
- [x] where do we fit keys
- [x] find spelling/grammar mistakes
- [x] responsiveness


developit:
Alrighty, I finally got around to finishing up my tweaks here. Here's what I changed:

- move all tutorial content to Markdown and create a couple widgets for defining `repl-initial` / `repl-final` and setup code blocks.
- expanded on the content in each chapter a fair bit, trying to explain even more about the fundamentals of things like VDOM trees, state updates, keys, refs, etc.
- split the Keys blurb out from the Side Effects chapter into a dedicated Keys chapter
- added automated solution checking for all chapters. The interactive solutions even require clicking!
- rebuilt the REPL's `<Runner>` component to run code in an `<iframe>` rather than directly in the page
    - I did this because I found it surprisingly easy to crash the tutorial page by deadlocking the render queue
    - code running in the tutorial and REPL now uses an independent copy of Preact and other libraries, and each change brings in fresh copies to prevent pollution across changes.
    - I also updated the REPL to fix share links, which are now Base64-encoded.

developit (approved):
I think we're good to go!

PR #821, opened 2022-02-05
Add section: set up Preact+TypeScript with CLI

drzamich:
Hi! I believe it would be beneficial for many beginners to include a section on how to set up a Preact project with TypeScript using Preact CLI.

I stumbled upon [TypeScript](https://preactjs.com/guide/v10/typescript) page today and was under impression that to set it up with Preact one needs to manually add `typescript` dependency and then play with compiler options. This resulted in much frustration, as it wasn't just as simple. Only after some digging I found out about dedicated typescript template of the CLI. I think adding such a section could save many people coming to Preact a lot of trouble.

### DIFF
--- a/content/en/guide/v10/typescript.md
+++ b/content/en/guide/v10/typescript.md
@@ -15,6 +15,15 @@ When you use Preact in a TypeScript-aware editor (like VSCode), you can benefit 
 
 ---
 
+## Setting up new project
+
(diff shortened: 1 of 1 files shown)


rschristian:
Hm, I wonder if that's more of a failing of our Preact-CLI docs than the Typescript docs? We're admittedly missing references to a few official templates on CLI's [Getting Started Page](https://preactjs.com/cli/getting-started), including the TS template, which certainly should be corrected and I'm hoping this would've put you on the right track faster?

In general, most sections don't contain references to CLI even when CLI comes with full support right out-of-the-box, like [Testing](https://preactjs.com/guide/v10/unit-testing-with-enzyme) and [SSR](https://preactjs.com/guide/v10/server-side-rendering), so I'd hesitate to add a CLI section to the TS docs.

rschristian:
Closing this out as we don't recommend CLI anymore

PR #832, opened 2022-03-01
docs: Copy 'skipLibCheck' note on 'Getting Started' to TS page

rschristian:
We have [this note on the 'Getting Started' page](https://preactjs.com/guide/v10/getting-started#typescript-preactcompat-configuration), but oddly it was missing from the actual Typescript page.

This is just a direct copy of that text.


rschristian:
Closing for #850

PR #833, opened 2022-03-03
Add banner to support Ukraine

marvinhagemeister:
Screenshot 2022-03-03 at 23 10 32


JoviDeCroock (approved):

ForsakenHarmony (approved):

hihanley:
This is not good. 
We all sympathize with the Ukrainian people, but preact, as an open source project, should not be exploited.
And this behavior will only make things worse, because they only want to hit Russia.
They only want to use Ukraine, to use our sympathy to achieve their goals.

Issue #834, opened 2022-03-04
Tutorial landing page pushes into the header


chakrakan:
Currently, the tutorial landing page instructions push into the header as seen below

#### Safari

<img width="1210" alt="Screen Shot 2022-03-03 at 9 18 56 PM" src="https://user-images.githubusercontent.com/57884127/156688475-cb87d61a-c934-403e-b003-b7229a51868d.png">

#### Chrome

<img width="1210" alt="Screen Shot 2022-03-03 at 9 19 12 PM" src="https://user-images.githubusercontent.com/57884127/156688546-510643d8-a5b0-497a-9605-e605a83adfee.png">

PR #835, opened 2022-03-04
fix: Update tutorial component style.module.less to lower the content height

chakrakan:
Addresses #834 

#### Before

Screen Shot 2022-03-03 at 9 19 12 PM

#### After

Screen Shot 2022-03-03 at 9 41 21 PM

### DIFF
--- a/src/style/variables.less
+++ b/src/style/variables.less
@@ -1,5 +1,4 @@
 @header-height: 3.5rem; // 56px
-@app-main-height: @header-height + 2.5rem; // 96px for banner
 @header-mobile-breakpoint: 50rem;
 @sidebar-width: 20rem;
 @sidebar-break: ~'(min-width: 900px)';
--- a/src/style/variables.less
+++ b/src/style/variables.less
@@ -1,4 +1,5 @@
-@header-height: 5rem; // 80px
+@header-height: 3.5rem; // 56px
+@app-main-height: @header-height + 2.5rem; // 96px for banner
 @header-mobile-breakpoint: 50rem;
 @sidebar-width: 20rem;
 @sidebar-break: ~'(min-width: 900px)';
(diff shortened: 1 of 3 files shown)


rschristian:
I think we need to correct this on all pages, not just the tutorial. The new banner height threw it off and covers some elements.

src/style/variables.less
```diff
@@ -1,4 +1,4 @@
-@header-height: 3.5rem; // 56px
+@header-height: 5rem; // 80px
```
rschristian, line 1: Definitely don't want to do this, it increases the header size significantly.

Instead, increase the padding on the main content:

https://github.com/preactjs/preact-www/blob/b04ccc87e480c950a8a55b2a376d24dea09dda36/src/style/index.less#L93

```diff
-padding-top: @header-height;
+padding-top: calc(@header-height + 2.5rem);
```

rschristian (commented):
Thanks for fixing this! Was just starting to look into it myself.

src/style/variables.less
chakrakan: updated!

src/style/variables.less
```diff
@@ -1,4 +1,5 @@
 @header-height: 3.5rem; // 56px
+@app-main-height: @header-height + 2.5rem; // 96px for banner
```
rschristian, line 2: nit: I don't think this needs to be extracted to a variable, as it's only used in 1 place.

I do like having the comment there for explaining the extra `2.5rem` though, good idea.

chakrakan:
This change should keep things consistent with the banner throughout `Tutorial` and `Guide` pages, which seemed to be the ones affected

### Guides 

Screen Shot 2022-03-04 at 1 30 34 AM

Screen Shot 2022-03-04 at 1 29 35 AM

src/style/variables.less
chakrakan: makes sense, I had pushed the change prior to seeing your comment, I'll use `calc` directly within `index.less`

rschristian: Sounds good, thanks!

rschristian (approved):
Thanks for catching and fixing this so quickly!

PR #836, opened 2022-03-04
fix: Corrects padding to avoid header overlap

rschristian:
Bah, missed this in #835 and only after I said we don't need to extract this out to a variable as it's only used in one spot.

Ensures header doesn't overlap content headings


JoviDeCroock (approved):

marvinhagemeister:
@rschristian Thanks for fixing this and apologizes for not checking that on my end!

PR #837, opened 2022-03-04
Chore/links

### LINKED ISSUE #818: Broken link on page "libraries & addons"
https://github.com/preactjs/preact-www/blob/1233db0c8a54e41092a35043af0852a80ff31fec/content/en/about/libraries-addons.md#L12

Two things:
 1. It should probably point to "**preactjs**/preact-www", instead of "**developit**/preact-www".
 2. The path is missing the language folder ("/content/**en**/about/libraries-addons.md"). This is probably a problem for every translation of this page.

rschristian:
Closes #818


JoviDeCroock (approved):
oh waw this is great ❤️

PR #841, opened 2022-03-14
chore(sponsors): add upload.io

JoviDeCroock:


marvinhagemeister (approved):
Looks good! Let's also add them to the "companies are using Preact" page

PR #842, opened 2022-03-27
Update index.js

hmzisb:


src/components/we-are-using/index.js
```diff
@@ -258,6 +258,11 @@ const companies = [
 		href: 'https://sqlframes.com/',
 		logo: 'sqlframes.svg'
 	}
+	{
+		name: 'BestForAndroid',
```
binyamin, line 262:
```suggestion
	},
	{
		name: 'BestForAndroid',
```

src/components/we-are-using/index.js
```diff
@@ -258,6 +258,11 @@ const companies = [
...
+		href: 'https://bestforandroid.com/',
+		logo: 'https://bestforandroid.com/wp-content/uploads/2020/03/cropped-bestforandroid-1.png'
+	}
```
binyamin, line 265: Please place the image in the `src/assets/we-are-using/` folder.

binyamin (changes requested):

binyamin:
@hmzisb This is a Wordpress site. What does this have to do with Preact?

marvinhagemeister (changes requested):
I'm afraid we have to close this PR as the site contains links to circumvent licensing of android apps, which is not legal in many countries. That's not something we want to be associated with for legal reasons.

PR #844, opened 2022-04-26
Add system group company to the section we-are-using.

MortezaMirjavadi:
Add system group company to the section of we-are-using.


marvinhagemeister:
@MortezaMirjavadi I can't find Preact in the source code for the website linked in the changes. Can you elaborate where the system group company is using Preact?

MortezaMirjavadi:
@marvinhagemeister The company is a big and have a lot of different products in fintech scope, those products aren't public. Preact is working on couple of apps with micro frontend architecture which is a core behind them.

rschristian:
Going to close out as stale.

If you still work at the company & want this to land, please clean up the PR so that it only changes what's needed.

PR #850, opened 2022-05-20
docs: Adds `paths` to recommendation for TS config

rschristian:
Re: https://github.com/preactjs/preact/issues/2748 & https://github.com/preactjs/preact/issues/3547, `"paths"` seemingly is becoming more and more required in order to solve type discrepancies between Preact and React.

Also duplicates the information from the 'Getting Started' page to the TS page, as the info was otherwise a bit fragmented which is less than ideal.


marvinhagemeister (approved):

JoviDeCroock (approved):
Nice! Good shout

PR #851, opened 2022-05-23
fix: use `findByText` on `screen`

eventualbuddha:
I assume this was left over from the previous API where `findByText` was destructured from the `render` return value.


rschristian (approved):
Nice catch, thanks!

PR #858, opened 2022-07-07
Forms.md: Consistently accesses value from event in examples.

processprocess:
Small update to make value accessing consistent across examples.


rschristian (approved):
Sorry this went unnoticed!

PR #860, opened 2022-07-10
Added preact-particles

matteobruni:


matteoums (approved):

rschristian:
Sorry for the delay, but going to close on out as the linked library isn't actually using Preact, it's just `preact/compat`. Users would get the same experience using the React version of the lib.

PR #867, opened 2022-08-04
refactor: Makes v8 warning banner sticky

rschristian:
Saw someone again linking to the v8 docs today and thought a slightly more aggressive banner is appropriate, especially with v11 coming up.

This simply makes the warning banner sticky and places it below the header while the user scrolls the page. Hopefully this should catch users following outdated links to where the warning might not be immediately visible, [like here](https://preactjs.com/guide/v8/differences-to-react/#whats-included).

Before             |  After
w/out sticky banner  |  w/ sticky banner

### DIFF
--- a/src/components/controllers/page/index.js
+++ b/src/components/controllers/page/index.js
@@ -173,7 +173,7 @@ export default function Page({ route, prev, next }, ctx) {
 				<div class={style.inner}>
 					{isDocPage(url) && +store.state.docVersion !== AVAILABLE_DOCS[0] && (
 						<div class={style.oldDocsWarning}>
-							You are viewing the documentation for an older version of Preact.
+							You are viewing the documentation for an older version of Preact.{' '}
 							<a href={docsUrl}>Switch to the current version →</a>
 						</div>
 					)}
--- a/src/style/markdown.less
+++ b/src/style/markdown.less
@@ -376,6 +376,9 @@
 		&:target {
 			padding-top: @header-and-banner-height;
 		}
+		content-region[name*='v8'] &:target {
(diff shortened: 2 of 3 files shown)


developit:
Good idea. I've added some suggestions in #868 to further increase visibility - this page in particular still shows up aggressively in Google Search.

JoviDeCroock (approved):

src/style/markdown.less
```diff
@@ -381,9 +372,13 @@
 	h4,
 	h5,
 	h6 {
+		// fix anchor target positioning to account for fixed header
 		&:target {
 			padding-top: @header-and-banner-height;
 		}
+		content-region[name*='v8'] &:target {
+			padding-top: calc(@header-and-banner-height + 3.25rem);
```
rschristian, line 380: This is the only location in which the aggregate height for all 3 (banner, header, and v8 warning) matters, so I chose not to extract it out to a var.

Lmk if this is too magic number-y.

src/components/controllers/page/index.js
```diff
@@ -173,8 +173,8 @@ export default function Page({ route, prev, next }, ctx) {
 				<div class={style.inner}>
 					{isDocPage(url) && +store.state.docVersion !== AVAILABLE_DOCS[0] && (
 						<div class={style.oldDocsWarning}>
-							You are viewing the documentation for an older version of Preact.
-							Switch to the <a href={docsUrl}>current version</a>.
+							You are viewing the documentation for an older version of Preact.{' '}
```
developit, line 176: Heh - just came back to add that space. Good catch.

developit (approved):

marvinhagemeister (approved):

PR #868, opened 2022-08-04
More aggressive v8 banner (additional suggestions)

developit:
Additional suggestions for #867.

### DIFF
--- a/src/style/markdown.less
+++ b/src/style/markdown.less
@@ -52,8 +52,8 @@
 			content: '';
 			display: block;
 			position: absolute;
-			height: @header-height;
-			margin: -@header-height 0 0;
+			height: @header-and-banner-height;
+			margin: -@header-and-banner-height 0 0;
 		}
 	}
 
--- a/src/components/controllers/page/index.js
+++ b/src/components/controllers/page/index.js
@@ -174,7 +174,7 @@ export default function Page({ route, prev, next }, ctx) {
 					{isDocPage(url) && +store.state.docVersion !== AVAILABLE_DOCS[0] && (
 						<div class={style.oldDocsWarning}>
 							You are viewing the documentation for an older version of Preact.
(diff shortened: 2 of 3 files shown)


JoviDeCroock (approved):

src/style/markdown.less
```diff
@@ -52,8 +52,8 @@
 			content: '';
 			display: block;
 			position: absolute;
-			height: @header-height;
-			margin: -@header-height 0 0;
+			height: @header-and-banner-height;
+			margin: -@header-and-banner-height 0 0;
```
rschristian, line 56: As far as I can tell this doesn't have any affect?

It certainly does _looks_ more correct to me though, not entirely sure why the existing setup works now that I'm looking at it.

developit: This only affects the offset scrolled to for headers when clicking anchor links. I think because the banner is only 50px we didn't notice.

rschristian: Maybe in theory, though I can't see an actual visual change?

We separately set the padding to 6rem (looks to be hard-coded, should be using `@header-and-banner-height`) so these two values should be fine to remove AFAICT.

developit: Where's the padding being set elsewhere? Can't find it

rschristian: Shouldn't have any visual effect. We already (separately_ set a `padding-top` to  `@header-and-banner-height`:

https://github.com/preactjs/preact-www/blob/7ebfaf38e3010990ad2f7805b684e0ded175deed/src/style/markdown.less#L378-L387

This is correct (where the `height` and `margin` values above are incorrect) which results in the working UI we have now.

This whole block is redundant now.

rschristian: Merged and pushed removal of that block. Preview should show it looking fine

PR #869, opened 2022-08-19
Adding Babbel to the Who's using Preact? page

mangs:
We've been using Preact as a team since 2019, and @developit suggested to me on Twitter that I add Babbel (the company I work at) to the "Who's using Preact?" page. So here I am. Thank you for the suggestion!


JoviDeCroock (approved):
Awesome, thank you!

mangs:
Thank you for the quick look!

PR #882, opened 2022-09-04
docs: add aliasing docs for vite

ppalone:
- [x] Added docs for aliasing React to Preact in Vite.

### DIFF
--- a/content/en/guide/v10/getting-started.md
+++ b/content/en/guide/v10/getting-started.md
@@ -241,6 +241,27 @@ export default {
 }
 ```
 
+#### Aliasing in Vite
+
+To alias in [Vite](https://vitejs.dev/), you'll need to add a package import alias to the `vite.config.js` file.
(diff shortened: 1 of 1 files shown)


ppalone:
Looks like alias to `preact/compact` is already added if we are using `@preact/preset-vite` https://github.com/preactjs/preset-vite/blob/main/src/index.ts#L156-L162.

rschristian:
I'd lean towards saying this is unnecessary as not only is there a Vite preset doing this (and other useful things) already, but it's included in the official Vite templates.

marvinhagemeister:
Thanks for making this PR! I tend to agree with @rschristian as we mainly added this section for bundlers where we don't have a preset for. In the future, I'm hoping that we can change our doc to have a general "tooling" section or something similar which lists how to best use Preact with tool XY.

ppalone:
@rschristian & @marvinhagemeister Thanks for your input. I'll close this PR.

PR #884, opened 2022-09-05
Opt-in into repl instead of always on

marvinhagemeister:
This PR changes the conditions on which we decide if we should add the `Run in REPL` link at the top right corner of a code snippet. Before this PR we'd always show it for every `js` or `jsx` code snippet, despite most of them being illustrative and not intended to be run. Basically, nearly all of our repl links were broken.

This PR changes that in that adding the repl link is an opt-in. It can be enabled by adding the `// --repl` comment in the first line.

```js
console.log("this code snippet WILL NOT have the 'Run in REPL' link");
```

vs

```js
// --repl
console.log("this code snippet WILL have the 'Run in REPL' link");
```

The `// --repl` comment is replaced by a new line when passed to the actual repl.

For most code snippets this is not enough though, since they require a little setup to be runable like adding imports. To do that, I've added two additional markers which are stripped for the snippet, but shown in the repl.

- `// --repl-before` - Code that is prepended before the code snippet
- `// --repl-after` - Code that is appended after the code snippet

With these features added we can also get rid of the magic "always call the first function" in repl, which caused me several infinite loops when playing around with our upcoming project in the repl.

### DIFF
--- a/content/en/guide/v10/components.md
+++ b/content/en/guide/v10/components.md
@@ -21,6 +21,9 @@ Functional components are plain functions that receive `props` as the first argu
 
 ```jsx
 // --repl
+import { render } from 'preact';
+
+// --repl-before
 function MyComponent(props) {
   return <div>My name is {props.name}.</div>;
 }
--- a/content/en/guide/v10/components.md
+++ b/content/en/guide/v10/components.md
@@ -45,6 +45,9 @@ Here we have a simple class component called `<Clock>` that displays the current
 
 ```jsx
 // --repl
+import { Component, render } from 'preact';
+
+// --repl-before
 class Clock extends Component {
 
   constructor() {
@@ -71,6 +74,8 @@ class Clock extends Component {
     return <span>{time}</span>;
   }
 }
+// --repl-after
+render(<Clock />, document.getElementById('app'));
 ```
 
 ### Lifecycle Methods
@@ -99,6 +104,8 @@ When an error is caught, we can use this lifecycle to react to any errors and di
 
 ```jsx
 // --repl
+import { Component, render } from 'preact';
+// --repl-before
 class Catcher extends Component {
   
   constructor() {
@@ -117,6 +124,8 @@ class Catcher extends Component {
     return props.children;
   }
 }
+// --repl-after
+render(<Catcher />, document.getElementById('app'));
 ```
 
 ## Fragments
--- a/content/en/guide/v10/components.md
+++ b/content/en/guide/v10/components.md
@@ -20,6 +20,7 @@ There are two kinds of components in Preact, which we'll talk about in this guid
 Functional components are plain functions that receive `props` as the first argument. The function name **must** start with an uppercase letter in order for them to work in JSX.
 
 ```jsx
+// --repl
 function MyComponent(props) {
   return <div>My name is {props.name}.</div>;
 }
@@ -40,6 +41,7 @@ Class components can have state and lifecycle methods. The latter are special me
 Here we have a simple class component called `<Clock>` that displays the current time:
 
 ```jsx
+// --repl
 class Clock extends Component {
 
   constructor() {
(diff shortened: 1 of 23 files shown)


content/en/guide/v10/components.md
```diff
@@ -20,6 +20,7 @@ There are two kinds of components in Preact, which we'll talk about in this guid
 Functional components are plain functions that receive `props` as the first argument. The function name **must** start with an uppercase letter in order for them to work in JSX.
 
 ```jsx
+// --repl
```
rschristian, line 23:
```suggestion
// --repl
import { render } from 'preact';

// --repl-before
```

content/en/guide/v10/components.md
```diff
@@ -40,6 +41,7 @@ Class components can have state and lifecycle methods. The latter are special me
 Here we have a simple class component called `<Clock>` that displays the current time:
 
 ```jsx
+// --repl
```
rschristian, line 44:
```suggestion
// --repl
import { Component, render } from 'preact';

// --repl-before
```

We'll also need the `-after` but GitHub is annoying and won't you suggest on unchanged lines:

```
// --repl-after
render(<Clock />, document.body);
```

content/en/guide/v10/components.md
```diff
@@ -93,6 +95,7 @@ There is one lifecycle method that deserves a special recognition and that is `c
 When an error is caught, we can use this lifecycle to react to any errors and display a nice error message or any other fallback content.
 
 ```jsx
+// --repl
```
rschristian, line 98:
```suggestion
// --repl
import { Component, render } from 'preact';

// --repl-before
```

Same as above

```
// --repl-after
render(<Clock />, document.body);
```

content/en/guide/v10/components.md
```diff
@@ -118,6 +121,7 @@ class Catcher extends Component {
 A `Fragment` allows you to return multiple elements at once. They solve the limitation of JSX where every "block" must have a single root element. You'll often encounter them in combination with lists, tables or with CSS flexbox where any intermediate element would otherwise affect styling.
 
 ```jsx
+// --repl
```
rschristian, line 124: Need to define `container` in this one, or just replace it with `document.body` I suppose?

rschristian (commented):
Ah, brilliant! I meant to look into this at some point but hadn't gotten around to it.

Tangentially related, but do we want to keep imports out of the code snippets going forward? While I'm happy for `--repl-before` to exist, I wonder if excluding imports from the snippets is making them less useful. I'm sure at least one user has skimmed the Hooks page, skipped over the intro and gotten stuck as they didn't realize hooks needed to be imported from `preact/hooks`.

Showing imports might provide useful transparency? Not sure.

marvinhagemeister:
@rschristian That's a very good point. I was unsure about this myself when going through the hooks documentation. My feeling is that it needs an overall in general and I didn't want to blow up this PR unnecessarily. I don't have a good feeling so far as to where the threshold lies after which a snippet should become runnable, so I kinda went a little on the conservative side for now.

rschristian:
Totally fair, makes sense!

I'll make a note and maybe run through if I get bored at some point.

content/en/guide/v10/components.md
marvinhagemeister: Good point, missed that. I kinda want to move folks away from `document.body`. Right now it kinda works because we sort-of make an effort to preserve existing DOM nodes in current Preact, but it always felt a bit risky to me to rely on that behaviour.

In our repl we always add a `<div id="app" />` with this PR to the DOM. Probably best to just switch to that.

rschristian: Oh, didn't know we added that.

I suggested `document.body` as it matches the others on the page. Can leave that for later too, rather than rewriting every sample we may show using it.

JoviDeCroock (approved):

rschristian (approved):
Huge improvement!

PR #887, opened 2022-09-06
fix: broken link

jsejcksn:
Fixed broken link to the CLI pre-rendering section from `hydrate()`. The unmodified translations did not appear to have the corresponding document/link.


rschristian (approved):
Thanks! We recently reshuffled some content and this moved. Will take a look now to see if we missed any others.

PR #889, opened 2022-09-06
Signals

marvinhagemeister:
🎉


content/en/guide/v10/signals.md
```diff
@@ -0,0 +1,529 @@
...
+```js
+const name = signal("Jane");
+const surname = signal("Doe");
+
+const fullName = computed(() => `${name.value} ${surname.value}`);
+```
+
+When creating computed signals within a component, use the hook variant: `useComputed(fn)`.
+
+### effect(fn)
```
JoviDeCroock, line 501: Should we include a chapter about effect in a component, this feels like a pit fall atm for me atleast

JoviDeCroock (approved):
YEAYA

PR #896, opened 2022-09-07
Update project-goals.md

quirico:
Fix spelling

### DIFF
--- a/content/en/about/project-goals.md
+++ b/content/en/about/project-goals.md
@@ -12,7 +12,7 @@ Preact aims to deliver on a few key goals:
 
 - **Performance:** Render quickly & efficiently
 - **Size:** Small size, lightweight _(approximately 3.5 kB)_
-- **Efficiency:** Effective memory usage _(avoiding GC thrash)_
+- **Efficiency:** Effective memory usage _(avoiding garbage collection trash)_
 - **Understandability:** Understanding the codebase should take no more than a few hours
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks for the PR 👍

content/en/about/project-goals.md
```diff
@@ -12,7 +12,7 @@ Preact aims to deliver on a few key goals:
 
 - **Performance:** Render quickly & efficiently
 - **Size:** Small size, lightweight _(approximately 3.5 kB)_
-- **Efficiency:** Effective memory usage _(avoiding GC thrash)_
+- **Efficiency:** Effective memory usage _(avoiding garbage collection trash)_
```
rschristian, line 15: Thrash is the right word here. "avoiding garbage collection trash" doesn't make much sense.

quirico: Sorry for this, not being a native speaker makes difference here. May I say that it still remains unclear to me, indeed?

rschristian: It's just something you'll need to learn then.

Thrashing, especially in relation to GC, is terminology in this field. While "thrash" may be a less common word, it's the correct one for this use case. See: https://en.wikipedia.org/wiki/Thrashing_(computer_science)

quirico: Right, @rschristian, thank you for the explanation and the link! :)

PR #900, opened 2022-09-11
(docs): introduce the `useId` hook

JoviDeCroock:


content/en/guide/v10/hooks.md
```diff
@@ -429,3 +429,34 @@ const App = props => {
 
 > If you've been using the class based component API in the past, then this hook is essentially an alternative to the [componentDidCatch](https://preactjs.com/guide/v10/whats-new/#componentdidcatch) lifecycle method.
 > This hook was introduced with Preact 10.2.0 .
+
+### useId
+
+This hook will generate a unique identifier for each invocation, these will be
+consistent across the server- and client-render. These id's can be used to
+have consistent and unique identifiers for dom-nodes, ...
```
marvinhagemeister, line 437: Nit: Tried to sneak in an example usage of where this is useful.

```suggestion
This hook will generate a unique identifier for each invocation and guarantees that these will be consistent when rendering both [on the server](/guide/v10/server-side-rendering) and the client. A common use case for consistent id's are forms, where `<label>`-elements use the [`for`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) attribute to associate them with a specific `<input>`-element. The `useId` hook isn't tied to just forms though and can be used whenever you need a unique id.
```

marvinhagemeister (approved):

content/en/guide/v10/hooks.md
```diff
@@ -429,3 +429,32 @@ const App = props => {
 
 > If you've been using the class based component API in the past, then this hook is essentially an alternative to the [componentDidCatch](https://preactjs.com/guide/v10/whats-new/#componentdidcatch) lifecycle method.
 > This hook was introduced with Preact 10.2.0 .
+
+### useId
+
+This hook will generate a unique identifier for each invocation and guarantees that these will be consistent when rendering both [on the server](/guide/v10/server-side-rendering) and the client. A common use case for consistent id's are forms, where `<label>`-elements use the [`for`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) attribute to associate them with a specific `<input>`-element. The `useId` hook isn't tied to just forms though and can be used whenever you need a unique id.
```
rschristian, line 435:
```suggestion
This hook will generate a unique identifier for each invocation and guarantees that these will be consistent when rendering both [on the server](/guide/v10/server-side-rendering) and the client. A common use case for consistent IDs are forms, where `<label>`-elements use the [`for`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) attribute to associate them with a specific `<input>`-element. The `useId` hook isn't tied to just forms though and can be used whenever you need a unique ID.
```

rschristian (approved):
Is this intended to be grouped with the side effect hooks?

Sounds a bit odd to me, but admittedly not too familiar with it or its common uses.

marvinhagemeister (approved):

PR #906, opened 2022-09-17
feat: Add additional language picker in header

### LINKED ISSUE #899: Proposal: Move language picker to the header
[translated]

> I found this after reading to the bottom.....
> 
> temp

Source: https://twitter.com/su8ru_/status/1568246589545938944

Having the language picker hidden away at the bottom of the page likely makes these situations occur more than they should. Moving to the header (which is always visible) should help this a lot.

rschristian:
Closes #899 

Adds additional language picker into the header. It doesn't replace as our social icons don't show up on mobile (not sure if that's intentional or not) and existing users might be confused if it just moved. Figured it probably doesn't do any harm if it is duplicated on desktop.

Here's what it looks like:

Image showing the language picker as an icon in the header. Image specifically shows the language picker "open", displaying possible language choices.


src/components/header/index.js
```diff
@@ -61,6 +62,7 @@ export default function Header() {
 								height="28"
 							/>
 						</a>
+						<NavItem />
```
rschristian, line 65: Reusing `NavItem` is probably less than ideal, at the very least from a naming perspective, but the prospect of rewriting/renaming everything in this component to be aware of non-link items was... less than appealing.

I hope that's not a problem, but maybe can find the time to rewrite everything if it will be.

marvinhagemeister: Agree, that's fine for now 👍

marvinhagemeister (approved):
This is fantastic 🙌

src/components/header/index.js
```diff
@@ -187,6 +198,40 @@ const NavLink = ({ to, isOpen, route, ...props }) => {
...
+	);
+
+	return (
+		<div
+			{...props}
+			data-open={isOpen}
+			class={cx(style.navGroup, style.translation)}
+		>
+			<button {...props} onClick={toggle} aria-haspopup aria-expanded={isOpen}>
+				<img src="/assets/i18n.svg" alt="Twitter" width="34" height="28" />
```
rschristian, line 218:
```suggestion
				<img src="/assets/i18n.svg" alt="Translate Page" width="34" height="28" />
```
Oops

PR #907, opened 2022-09-17
chore: Bump preact-cli, remove testing, and other small misc maintainence items

rschristian:
Need to slowly go through everything still just to ensure nothing's broken, but first pass looks alright.

Summary:
 - Bumped `preact-cli` to latest. Gets a newer version of `webpack-dev-server`, but nothing too noteworthy there.
 - Removed old testing setup. Was unused & good portion of the project's total dependencies.
 - Disabled Critical CSS generation.
   - We were using a newer version of Critters here, which has a [duplication issue](https://github.com/GoogleChromeLabs/critters/issues/88). However, reverting to a previous version increases build times by ~50% (~45s -> 66s). `cssnano` was struggling. See the brief conversation below, but I don't think it's a huge benefit anyways.
 - Removed some other misc deps that we didn't need and some fixes in the CLI config that were years out-of-date (and no longer needed). 
 - Removed `size-plugin-ssr` tracking. It's quite meaningless due to the lack of code splitting in ssr-bundles and we haven't touched it in years.

I don't mind separating some of these out if it helps. Kinda got carried away with "If I'm doing x I may as well do y while I'm at it...".


marvinhagemeister:
@rschristian These days I'm in favor of dropping critical CSS all together. I don't think it gains us much in perf.

Agree about the testing suite. The few tests we have aren't really useful, and if we want to really invest into testing, then we should use `playwright` instead. So feel free to drop the existing ones.

rschristian:
Fair, I'm kind of in the same boat after reading [that article](https://csswizardry.com/2022/09/critical-css-not-so-fast/) that made the rounds last week and doing some of my own testing. Hadn't explored it too much prior but indeed seems to be rather lacking in benefit. Especially here where it seems `cssnano` is struggling.

marvinhagemeister (approved):

package.json
```diff
@@ -76,29 +68,20 @@
...
-    "dlv": "^1.1.3",
     "eslint": "^6.8.0",
     "eslint-config-developit": "^1.1.1",
     "husky": "^4.0.0",
-    "identity-obj-proxy": "^3.0.0",
-    "jest": "^26.6.3",
-    "jest-puppeteer": "^4.4.0",
     "jsdom": "^15.2.1",
-    "less-loader": "^5.0.0",
+    "less-loader": "^7.3.0",
```
rschristian, line 76: Was the last of the peer-dep issues, so no longer need `--legacy-peer-deps` upon installing with newer NPM versions.

PR #908, opened 2022-09-17
fix: Correct alt text on translation/i18n icon

rschristian:
Silly mistake from a lazy copy/paste, sorry.


marvinhagemeister (approved):
All good, happens to everyone (myself included) 👍

PR #913, opened 2022-09-21
Add Zeplin to companies using Preact

merih:
We're using Preact since 2017!


developit (approved):
Awesome! I totally knew that and should have thought to reach out sooner!

PR #915, opened 2022-09-22
Add "Signal Boosting" blog post

jviide:
This pull request adds the Signal Boosting blog post, outlining what we did to optimize Signals Core v1.2.0.

I assumed blog posts should be listed in reverse order, but I'm not 100% sure 🙂


marvinhagemeister:
FYI: We'll time the publishing later today for maximum impact 👍

content/en/blog/signal-boosting.md
```diff
@@ -0,0 +1,290 @@
+---
+title: Signal Boosting
+date: 2022-09-24
+---
+
+The new release of Preact Signals brings significant performance updates to the underpinnings of the reactive system. Read on to learn what kinds of tricks we employed to make this happen.
```
JoviDeCroock, line 6:
```suggestion
The new release of Preact Signals brings significant performance updates to the foundations of the reactive system. Read on to learn what kinds of tricks we employed to make this happen.
```

Wording suggestion, could be me not being familiar with "underpinngs" as expression so feel free to ignore

content/en/blog/signal-boosting.md
```diff
@@ -0,0 +1,290 @@
...
+});
+```
+
+The compute function given to `computed(...)` won't run immediately, so the `"I'm computing!` text won't get printed to the console. That's because computed signals are evaluated _lazily_, i.e. when their values are read.
+
+```ts
+console.log(c.value); // Console: Hello World
+```
+
+Computed values are also _cached_. A compute function can potentially be expensive, so we want to recompute only when it matters. A running compute function tracks which signal values are actually accessed during its run. When none of the signals have changed, then we can skip recomputation. In the above example, we can just reuse the previously calculated `c.value` indefinitely until either `a` or `b` changes. After that the compute function needs to be evaluated again on next read. Facilitating this _dependency tracking_ is the reason why we need to wrap the primitive values into signals in the first place.
```
JoviDeCroock, line 66: Don't know if it's worth it but we could couple back to a smart `memoize` function that knows when to recalculate

JoviDeCroock (approved):
Amazing, love the image with the sources and targets 😄

content/en/blog/signal-boosting.md
```diff
@@ -0,0 +1,290 @@
+---
+title: Signal Boosting
+date: 2022-09-24
+---
```
rschristian, line 4:
```suggestion
authors:
  - Joachim Viide
---
```
[Marvin's earlier PR](#916) adds support for specifying authors and [he's conveniently already added your details](https://github.com/preactjs/preact-www/blob/69398eda85599551cfc04df82889259a44d05f22/src/config.json#L512-L515).

Apologies if anyone else also contributed, I'm unaware.

content/en/blog/signal-boosting.md
jviide: Thanks, that's good 👍

content/en/blog/signal-boosting.md
jviide: Thanks! 👍

content/en/blog/signal-boosting.md
jviide: Great suggestion 👍 Mentioning `memoize` would totally give some context for many readers. I tried modifying the text based on your suggestion, but couldn't figure out a way smooth way to do it yet.

marvinhagemeister (approved):

PR #932, opened 2022-10-06
Update _redirects

bosung90:
Fix redirect to include switching-to-preact


rschristian:
Looks like we actually already have this redirect:

https://github.com/preactjs/preact-www/blob/218b51104cd24819eff08b832eeb79471ebdee63/src/_redirects#L3

Therefore closing this out.

bosung90:
Ah my bad, yes I found the link somewhere in the guide

PR #933, opened 2022-10-08
Added TR language support v10 documentation beginner level support.

Kripteks:


rschristian (changes requested):
Thanks for the PR!

It looks like you've copied the English directory though, only translating a few of the files. All English files will need to be removed.

The way our system works is that it tries to find the translated version of a page when a user navigates to it. If it cannot, it will fallback to the English translation and alert the user that no translation for the page could be found. Therefore, `content/tr` should only contain translated documents. If something's missing, it'll fallback to English automatically.

Kripteks:
I will make a detailed translation and publish it as soon as possible.

rschristian:
Just to clarify, I'm happy to merge any amount of translations -- every little bit helps! We certainly don't expect you to translate everything, as that'd be a ton of work!

But for translated content, we only want the actual translations included. We don't want English documents copied over, as that breaks our detection.

PR #934, opened 2022-10-09
docs: update preact-signal-store to @deepsignal/preact in libraries-addons.md

EthanStandel:
I'm the maintainer of the library previously named `preact-signal-store`, but to support React I recently deprecated the library under that name so that I could create scoped packages like the main Signals repo.

This PR is just updating the reference to the deprecated repo to have the name of the new repo.


rschristian (approved):
Thanks!

PR #935, opened 2022-10-14
Add missing `waitFor` import

christian-bromann:
minor example fix


marvinhagemeister (approved):
Great catch! Thanks for making a PR 👍

PR #937, opened 2022-10-24
Typo fix

VofSwords:
Perhaps this is not a typo, but a feature. Otherwise why is it here for almost a year


rschristian (approved):
Not a feature, just a mistake that's gone unnoticed.

Appreciate you taking the time to PR this!

PR #939, opened 2022-11-10
Update Simplified Chinese translation

Byacrya:
This pull request translates multiple docs into Simplified Chinese.


marvinhagemeister (approved):
Woa, this is awesome! Thanks you so much for the PR 🙌

PR #139, opened 2017-07-09
WIP: Arabic content translation

zouhir:
That's still a [WIP]

[x] 404.md (100% done)
[] index.md (85% done)

Note to self: Open an issue so <p> and <h*> tags can have `RTL` direction if that doesn't exist.

PR #943, opened 2022-11-17
RTL - Persian Language Support

iarata:
Currently, Index and 404 are done.
The CSS styling supporting the RTL language has been added (tested on the above pages).

Since the site doesn't support RTL, at this moment I think it would be better to have a separate branch for RTL support until all pages or a v.1 support is released.

I also saw the #139 but only translation and no CSS support.


iarata:
A separate `rtl.less` file is created to write the RTL support there.
For the `ContentRegion` function if the language is Persian, the RTL class name is added based on the `props.name`: blog-rtl, getting-started-rtl, ...

PR #947, opened 2022-12-01
docs: Remove Preact-CLI as a recommended tool

rschristian:
Re: https://github.com/preactjs/preact/pull/3816

A recommendation for a different tool would be better but I'm not sure if we have any consensus there on what we want to push. I feel bad every time someone new to Preact picks up CLI though, as it's not in a good way and will influence their view of the ecosystem rather negatively in all likelihood.


rschristian:
Superseded by #1002

PR #950, opened 2022-12-27
fix: Correct language picker dropdown arrow on light mode

rschristian:
It'll turn black if the user browses in light mode, which isn't right.

The `vertical-align` is unused; a leftover from when I was testing out various methods I suppose.


developit (approved):

developit:
hah, I thought this was a select dropdown styling issue. good catch.

PR #951, opened 2022-12-30
feat: Prerender current Preact release in header

rschristian:
The current Preact release number flashes in the header due to us only prerendering our fallback version of 10.6.4. This was somewhat bothering me, so this PR reduces the chances of it occurring. It'll only flash when there has been a Preact release without a rebuild of these docs.

Also removed `src/ssr.js` & `src/promise-polyfill.js`; the former is unused and the latter has been unnecessary since a CLI update (per the comment in the header of that file) long ago.


src/lib/github.js
```diff
@@ -21,8 +21,4 @@ export const fetchRelease = repo =>
 		.then(d => ({
 			version: d.version || '??',
 			url: d.url || 'https://github.com/preactjs/preact'
-		}))
-		.catch(() => ({
-			url: '#',
-			version: '??'
```
rschristian, line 27: Not sure I understand the original idea with this. We have a fallback version and link for a release, so I don't think there's any situation in which we'd want to fall all the way back to displaying `v??` ?

src/prerender.js
```diff
@@ -18,20 +18,17 @@ const FRONT_MATTER_REG = /^\s*---\n\s*([\s\S]*?)\s*\n---\n/i;
 // https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup
 const MAX_DESCRIPTION_LENGTH = 200;
 
-if (process.env.PRERENDER_HOME) {
```
rschristian, line 21: AFAIK this is unused? The blame points to a ["wip" commit 3 years ago](https://github.com/preactjs/preact-www/commit/a8957534f632c8ad32ea069c9063828a027bd354) so I went out on a limb and guessed it's not needed.

marvinhagemeister (approved):

src/lib/github.js
marvinhagemeister: I think it was mainly for local dev from the early days

rschristian: Fair enough, admittedly I didn't dig into the blame on that one. Might've predated the fallback version / link.

PR #957, opened 2023-01-19
Changes `target` to `currentTarget`.

MicahZoltu:
I believe that in most cases `currentTarget` is what the user wants, rather than `target`, and `currentTarget` also type checks much better than `target` in preact TS.  If someone attaches a handler to an element, it is very likely that they want to read data off of that element, not some child element that happened to receive the actual click/input.

In the case of these two examples, either `target` or `currentTarget` will work, but the code is incorrect if the element in question has any children that don't have a `value` property, and most likely the `value` on such a child wouldn't be the one you intended to pick up on.

----

If this change is generally accepted, I may submit another PR that tries to make this same correction in a number of other locations in the documentation, but I wanted to get some initial feedback on this smaller PR first.


MicahZoltu:
I am pretty busy these days, so I would recommend merging this as-is if you find it acceptable and not waiting on me to submit a follow-up PR if you have a hankering to fix the other places.

rschristian:
No worries, will address it myself tomorrow.

Sorry again for it slipping through the cracks.

PR #959, opened 2023-01-21
Updates signals to latest.

### LINKED ISSUE #958: Update @preact/signals.
This project uses `@preact/signals` version `1.0.0`, which I believe means the REPL uses `1.0.0`.  `useSignalEffect` does not appear to be present in `1.0.0`, which means it cannot be used in the REPL.

What is the process for updating a dependency on this project?  Should I just submit a PR that updates package.json and rebuilds the lockfile?

MicahZoltu:
Fixes #958

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "preact-www",
       "license": "MIT",
       "dependencies": {
-        "@preact/signals": "^1.0.0",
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Awesome, thanks a bunch for the PR 👍

PR #968, opened 2023-02-27
Fix scroll to anchor when navigating pages

kidonng:
As I submitted #967, I realized the page would scroll to an anchor only if the links are opened in a new page. When navigating pages normally, links with anchor doesn't work as expected.

This pull request tries to fix the above issue, and I welcome any better solution.


src/components/controllers/page/index.js
```diff
@@ -161,6 +161,17 @@ export default function Page({ route, prev, next }, ctx) {
 	const canEdit = showTitle && current != '404' && current !== '/blog';
 	const hasSidebar = meta.toc !== false && isDocPage(url);
 
+	useEffect(() => {
+		if (location.hash) {
+			const anchor = document.querySelector(location.hash);
+			if (anchor) {
+				// Do not use scrollIntoView as it will cause
+				// the heading to be covered by the header
+				scrollTo({ top: anchor.offsetTop });
```
developit, line 170:
```suggestion
				scrollTo(0, anchor.getBoundingClientRect().top);
```

rschristian: Looking at this again, `getBoundingClientRect().top` would get the top of the wrapping element and would need to be inverted (that will return a negative when off the bottom of the screen).

rschristian:
Sorry for the long delay here.

The remaining trouble here is with the table of contents, which is simultaneously generated (and will further affect the positioning). We'll need to wait for that to be complete before scrolling the user.

src/components/code-block/index.js
```diff
@@ -135,7 +135,14 @@ const CodeBlock = props => {
 		)[1];
 		const firstChild = getChild(child.props);
 		const code = String(firstChild || '').replace(/(^\s+|\s+$)/g, '');
-		return <HighlightedCodeBlock {...props} code={code} lang={lang} />;
+		return (
+			<HighlightedCodeBlock
+				{...props}
+				code={code}
+				lang={lang}
+				key={lang + '\n' + code}
```
rschristian, line 143: Same ID we use for the highlight cache -- without the forced remount, we'll have a stale highlight value from the previous page's code block for a render (thereby throwing our positioning off)

rschristian (approved):
Sorry again for the delay here!

PR #970, opened 2023-03-05
Typo: fix missing word on Signals page

tomodwyer:
I was reading the signals documentation and spotted what I believe to be a missing word. I believe the word missing is "use".


rschristian (approved):
Nice catch, thanks!

Issue #975, opened 2023-03-13
Tutorial 05-refs uses incorrect HTML attribute


masto:
I'm not an expert, but I noticed this wasn't working:

![image](https://user-images.githubusercontent.com/645516/224595443-229aad19-db57-4313-93f8-89a0a9ed7213.jpg)

and I can't find any reference to `initialValue` being a thing.

rschristian:
Fixed in #976

PR #976, opened 2023-03-13
Correct attribute for initial value of input field

masto:
Fix for #975

### DIFF
--- a/content/en/tutorial/05-refs.md
+++ b/content/en/tutorial/05-refs.md
@@ -177,7 +177,7 @@ function App() {
 
   return (
     <div>
-      <input initialValue="Hello World!" />
+      <input value="Hello World!" />
       <button onClick={onClick}>Focus input</button>
     </div>
   );
@@ -199,7 +199,7 @@ function App() {
 
   return (
     <div>
-      <input ref={input} initialValue="Hello World!" />
+      <input ref={input} value="Hello World!" />
       <button onClick={onClick}>Focus input</button>
     </div>
   );
--- a/content/en/tutorial/05-refs.md
+++ b/content/en/tutorial/05-refs.md
@@ -177,7 +177,7 @@ function App() {
 
   return (
     <div>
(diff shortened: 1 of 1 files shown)


content/en/tutorial/05-refs.md
```diff
@@ -177,7 +177,7 @@ function App() {
 
   return (
     <div>
-      <input initialValue="Hello World!" />
+      <input value="Hello World!" />
```
rschristian, line 180:
```suggestion
      <input defaultValue="Hello World!" />
```

No real difference in this example, but I believe `defaultValue` is what they were mentally reaching for.

content/en/tutorial/05-refs.md
```diff
@@ -199,7 +199,7 @@ function App() {
 
   return (
     <div>
-      <input ref={input} initialValue="Hello World!" />
+      <input ref={input} value="Hello World!" />
```
rschristian, line 202:
```suggestion
      <input ref={input} defaultValue="Hello World!" />
```

rschristian (approved):

content/en/tutorial/05-refs.md
masto: Thanks. My first guess was defaultValue (and it appeared to work), but then I checked https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attributes and didn't see it there, so I thought that might be some kind of unofficial behavior. But I see it's listed on https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement, so I guess the omission there was a mistake.

rschristian: `defaultValue` is a property, rather than an attribute. The distinction is the interface, as attributes are an HTML thing while properties are JS(X). It therefore gives access to a few things in an HTML-like interface that wouldn't actually work in HTML.

rschristian:
Thank you so much for taking the time to create issues and PRs, we really appreciate it!

PR #980, opened 2023-03-21
Adding `react-dom/client` to aliasing instructions

KaeruCT:
I had to add this to fix my jest setup, since it was importing the real `react-dom/client` module instead of `preact/compat/client`. I think this started happening from React 18 onwards.

### DIFF
--- a/content/en/guide/v10/getting-started.md
+++ b/content/en/guide/v10/getting-started.md
@@ -148,7 +148,8 @@ const config = {
     "alias": { 
       "react": "preact/compat",
       "react-dom/test-utils": "preact/test-utils",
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/getting-started.md
```diff
@@ -219,6 +222,7 @@ These rewrites are configured using regular expressions in your Jest configurati
   "moduleNameMapper": {
     "^react$": "preact/compat",
     "^react-dom/test-utils$": "preact/test-utils",
+    "^react-dom/client$": "preact/compat/client",
```
rschristian, line 225: Is this an issue for any tool outside of Jest? The regex pattern(s) we list here would obviously cause issues.

PR #983, opened 2023-04-05
docs: update React version compatibility

dijonkitchen:


content/en/guide/v10/differences-to-react.md
```diff
@@ -33,7 +33,7 @@ Another notable difference is that Preact follows the DOM specification more clo
 
 For both preact and [preact/compat], version compatibility is measured against the _current_ and _previous_ major releases of React. When new features are announced by the React team, they may be added to Preact's core if it makes sense given the [Project Goals]. This is a fairly democratic process, constantly evolving through discussion and decisions made in the open, using issues and pull requests.
 
-> Thus, the website and documentation reflect React `16.x` and `15.x` when discussing compatibility or making comparisons.
+> Thus, the website and documentation reflect React `18.x` and `17.x` when discussing compatibility or making comparisons.
```
rschristian, line 36: I don't think we can (safely) put `18.x` without any further qualifiers, as we are still missing some pieces of React 18 compat. [`preact/compat` still fakes being React v17.0.2](https://github.com/preactjs/preact/blob/90042a1a012855ef975309485534595705c777c5/compat/src/index.js#L38).

Maybe something like this instead?

```suggestion
> Thus, the website and documentation reflect React `15.x` through `17.x` when discussing compatibility or making comparisons.
```

While we do implement some React 18 features, this saves users from expecting full compatibility as we don't (yet) list the differences anywhere.

andrewiggins (approved):
Thanks y'all!

PR #985, opened 2023-04-14
fix: Correct rts imports in pretty SSR examples

### LINKED ISSUE #984: server-side-rendering
[server-side-rendering.md](https://github.com/preactjs/preact-www/blob/master/content/en/guide/v10/server-side-rendering.md) contains a section about pretty mode. But I think support for pretty mode has been removed?
When I look at the source code of preact-render-to-string there doesn't seem to be a third parameter for `renderToString`

rschristian:
Closes #984


marvinhagemeister (approved):
Good catch! I think we forgot to update those 👀

PR #986, opened 2023-04-14
feat: Show edit page link on tutorial pages

rschristian:


src/components/controllers/page/index.js
```diff
@@ -158,7 +158,7 @@ export default function Page({ route, prev, next }, ctx) {
 	// "current" is the currently *displayed* page ID.
 
 	const showTitle = current != 'index' && meta.show_title !== false;
-	const canEdit = showTitle && current != '404' && current !== '/blog';
```
rschristian, line 161: @marvinhagemeister Looks like this was done when you created the blog originally, do you remember the reasoning behind it?

Currently it doesn't work, as the page is just `blog`, without the leading `/`, but I don't necessarily see any reason why it should not be editable; in fact, someone's been kind enough already to add a [Chinese translation](https://github.com/preactjs/preact-www/blob/master/content/zh/blog.md).

src/components/controllers/tutorial/index.js
```diff
@@ -284,7 +286,13 @@ function TutorialView({
...
 			>
 				<div class={style.tutorialWindow} ref={content}>
-					<h1 class={style.title}>{title}</h1>
+					<Hydrator
+						boot={!loading}
+						component={EditThisPage}
+						show={title != 'Error'}
+						isFallback={page.isFallback}
+					/>
+					<h1>{title}</h1>
```
rschristian, line 295: `style.title` doesn't exist, hence the removal, in case that raises an eyebrow.

src/components/controllers/tutorial/index.js
```diff
@@ -284,7 +285,8 @@ function TutorialView({
 				}
 			>
 				<div class={style.tutorialWindow} ref={content}>
-					<h1 class={style.title}>{title}</h1>
+					<EditThisPage show={title != 'Error'} isFallback={page.isFallback} />
+					<h1>{title}</h1>
```
rschristian, line 289: `style.title` doesn't exist, hence the removal, in case that raises an eyebrow.

JoviDeCroock (approved):

src/components/controllers/page/index.js
marvinhagemeister: @rschristian Apologies for seeing the mention so late. I think originally when prototyping the blog we had nothing that could be translated on that page. Might have changed in a newer iteration so I likely just forgot to update this part.

rschristian: No worries!

And fair enough, just wanted to check to make sure I wasn't overlooking anything.

PR #987, opened 2023-04-14
fix: Minor styling issues in header

rschristian:
When header items with dropdown menus were opened, the drop down arrow would stay white, resulting in it disappearing:

Before             |  After
white-on-white drop-down arrow  |  black-on-white drop-down arrow

### DIFF
--- a/src/components/header/index.js
+++ b/src/components/header/index.js
@@ -65,9 +65,11 @@ export default function Header() {
 								height="28"
 							/>
 						</a>
-						<NavItem />
 					</div>
-					<Hamburger open={open} onClick={toggle} />
+					<div class={style.translation}>
+						<NavMenu language />
+					</div>
+					<HamburgerMenu open={open} onClick={toggle} />
 				</div>
 			</div>
 			<Corner />
@@ -75,8 +77,7 @@ export default function Header() {
 	);
 }
 
-// hamburger menu
-const Hamburger = ({ open, ...props }) => (
+const HamburgerMenu = ({ open, ...props }) => (
 	<div class={style.hamburger} open={open} {...props}>
 		<div class={style.hb1} />
 		<div class={style.hb2} />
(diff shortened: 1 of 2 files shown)


developit:
@rschristian I totally agree the language picker should be wider, but I think we need to adjust some other spacing a tiny bit - the wider one is making the REPL link drop out of the header at some screen widths:


https://user-images.githubusercontent.com/105127/232235918-a837742d-72d4-4dc6-8b54-912c5df4f510.mov

rschristian:
Oops, missed that. Will take a look

rschristian:
Alrighty, I just reverted the change to the translation button padding, it (and the header in general) needs a bigger time investment to fix.

src/components/header/style.module.less
```diff
@@ -389,6 +388,7 @@
 	height: 100%;
 	align-items: center;
 	border: none;
+	color: #eee;
```
rschristian, line 391: Text in the nav area is `#eee`, the social links are `#f0f0f0`, and the dropdown arrows were `#fff`.

I just picked the nav link text to match

rschristian:
Alrighty, I think this setup should work a bit better. No visual issues that I can see, and I've refactored some of the awkward (IMO) bits of the translation button setup. Should be less duplication of CSS and markup now, hopefully.

src/components/header/style.module.less
```diff
@@ -354,10 +354,17 @@
...
+
+.translation button {
+	padding: 0 0.6rem 0 0.5rem;
+}
+
 .socialItem,
-.translation {
+.translation button {
 	display: flex;
-	padding: 0 0.5rem;
```
rschristian, line 360: Moved the padding to the search box instead, made things slightly cleaner.

src/components/header/style.module.less
```diff
@@ -354,10 +354,17 @@
 	}
 }
 
+.socialItem {
+	padding: 0 0.5rem;
+}
+
+.translation button {
+	padding: 0 0.6rem 0 0.5rem;
```
rschristian, line 362: 0.5 is the standard padding for the social icons, whilst 0.6 is what the 'About' dropdown menu uses.

Therefore we use 0.5 on the left to keep it a bit tighter/more consistent with the social icons, but 0.6 on the right to give the dropdown arrow more breathing room.

src/components/header/index.js
```diff
@@ -87,25 +88,31 @@ const Hamburger = ({ open, ...props }) => (
...
+					class={cx(
+						route.class,
+						(pathMatchesRoute(current, route) ||
+							(route.content === 'guide' && /^\/guide\//.test(current)) ||
+							(route.content === 'blog' && /^\/blog\//.test(current))) &&
+							style.current
+					)}
+				/>
+			)
+		)}
```
rschristian, line 110: I thought it made a bit more sense to switch `NavItem` to `NavMenu`, and avoid passing `NavLinks` directly through.

marvinhagemeister (approved):

PR #989, opened 2023-04-15
fix: Mobile content width on tutorial

rschristian:
Before             |  After
Before, with horizontal scroll |  After, no horizontal scroll

Might be opinion, but I think it's a bit easier to read without the horizontal scroll if we can help it.

### DIFF
--- a/src/components/controllers/tutorial/style.module.less
+++ b/src/components/controllers/tutorial/style.module.less
@@ -55,7 +55,6 @@
 			background: var(--color-sidebar-bg);
 			color: var(--color-text);
 			display: flex;
-			flex: 1 1 var(--x, 50%);
 			flex-direction: column;
 			padding: 12px;
 			transition: opacity 1s ease;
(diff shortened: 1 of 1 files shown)


src/components/controllers/tutorial/style.module.less
```diff
@@ -55,7 +55,6 @@
 			background: var(--color-sidebar-bg);
 			color: var(--color-text);
 			display: flex;
-			flex: 1 1 var(--x, 50%);
```
rschristian, line 58: I noticed and removed a fair handful of unused flex rules. None of these elements are ever flex items as far as I can see, so I'm guessing they're just left overs.

JoviDeCroock (approved):

src/components/controllers/tutorial/style.module.less
developit: also me fumbling with Flexbox :P

developit (approved):

PR #990, opened 2023-04-16
refactor: Create a more prominent TS-subsection in guide

rschristian:
I'm thinking it might be better to move TS out of 'Advanced' and into a more prominent subsection -- no longer is TS really seen as an advanced feature, but table stakes.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -414,6 +414,25 @@
 		},
 		{
 			"name": {
+				"en": "TypeScript"
+			},
+			"routes": [
+				{
+					"path": "/guide/v10/configuring-typescript",
+					"name": {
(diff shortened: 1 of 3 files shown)


src/config.json
```diff
@@ -382,6 +382,55 @@
 				}
 			]
 		},
+		{
+			"name": {
+				"en": "React compatibility",
+				"zh": "React 兼容性"
+			},
```
rschristian, line 389: I moved this section up as compatibility, for better or worse, is a major part of our ecosystem and draw. Most users are probably going to want to read that right after intro + essentials, debugging can get bumped down a slot.

PR #996, opened 2023-05-03
fix: weird look

JajarGaming:
Looks pretty weird in the pages `toc`/Table of Contents: 

brave_BRjODqZaCn


rschristian (approved):
Good spot, thanks!

I guess we don't use backticks in any other header by the looks of it.

PR #999, opened 2023-06-08
Add support for translation authors

marvinhagemeister:
Best explained via a picture:

Screenshot 2023-06-08 at 21 23 56


rschristian (approved):
Looks great, love the giving attribution to the translator!

Not to put the cart before the horse, but I wonder if (later) we can do something along these lines on all docs pages, give a little bit more of a shout-out to the translators we depend on.

marvinhagemeister:
Great idea! We should definitely give more attribution to translators 👍

PR #1011, opened 2023-09-10
docs: Add note on `render()` instantiation

### LINKED ISSUE #974: Somewhat confusing introduction to Components at the start of the tutorial
https://github.com/preactjs/preact-www/blob/master/content/en/tutorial/01-vdom.md, under "Detour: Components", there is an example that says "[...] we don't need to write that last `render()` line anymore".

But if you actually try that example in the REPL, it does not work. In fact, the default code block loaded up ends with a `render()` line:

`render(<App />, document.getElementById("app"));`

As I am brand new to Preact, I found this discrepancy between the statement above and the actual behavior confusing. I expect it will make more sense later when there is presumably a full treatment of Components, but it might be better not to have any head-scratchers on page 1 of the tutorial.

### LINKED ISSUE #1010: Perhaps copy `render()`'s invoking note to the first page of the tutorial
https://github.com/preactjs/preact/issues/4120

A few people trip up on this rather important detail with the info being somewhat buried in the API docs.

rschristian:
Closes #974 
Closes #1010

### DIFF
--- a/content/en/tutorial/01-vdom.md
+++ b/content/en/tutorial/01-vdom.md
@@ -140,8 +140,7 @@ _two_ types of Virtual DOM elements. The other type is a Component, which is
 a Virtual DOM element where the type is a function instead of a string like `p`.
 
 Components are the building blocks of Virtual DOM applications. For now, we'll
-create a very simple component by moving our JSX into a function, which will be
-rendered for us so we don't need to write that last `render()` line anymore:
+create a very simple component by moving our JSX into a function:
 
 ```jsx
 import { createElement } from 'preact';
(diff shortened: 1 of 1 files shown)


content/en/tutorial/01-vdom.md
```diff
@@ -140,8 +140,7 @@ _two_ types of Virtual DOM elements. The other type is a Component, which is
 a Virtual DOM element where the type is a function instead of a string like `p`.
 
 Components are the building blocks of Virtual DOM applications. For now, we'll
-create a very simple component by moving our JSX into a function, which will be
-rendered for us so we don't need to write that last `render()` line anymore:
+create a very simple component by moving our JSX into a function:
```
rschristian, line 143: I get what this was going for, but I think it will be quite confusing for new users (as #974 rightly brings up).

content/en/tutorial/01-vdom.md
```diff
@@ -185,26 +202,26 @@ useResult(function(result) {
 
 
 ```jsx:repl-initial
-import { render } from 'preact';
+import { createElement, render } from 'preact';
```
rschristian, line 205: May or may not throw users off, but I figure it might help to be a bit consistent w/ JSX and importing `h`/`createElement`. Our other snippets on this page do import it.

content/en/tutorial/01-vdom.md
```diff
@@ -151,6 +150,24 @@ export default function App() {
...
+```jsx
+const App = () => <div>foo</div>;
+
+// DON'T: Invoking components directly breaks hooks and update ordering:
+render(App(), rootElement); // ERROR
+render(App, rootElement); // ERROR
+
+// DO: Passing components using createElement() or JSX allows Preact to render correctly:
+render(createElement(App), rootElement); // success
+render(<App />, rootElement); // success
```
rschristian, line 170: Pretty much directly copied from the [API reference](https://preactjs.com/guide/v10/api-reference#render) but with a switch to `createElement` as we use it, rather than `h`, on this page.

JoviDeCroock (approved):

marvinhagemeister (approved):
I like this!

PR #1012, opened 2023-09-11
feat: Add fragment links for easy linking of page subsections

rschristian:
Makes for an easier time of linking to subsections as users (likely us, most of the time) won't have to grab the link by first using the table of contents.

Limited to the guide at the moment; could support the tutorial, but would need a bit different styling as we can't get away with the negative margin there.


src/lib/content.js
```diff
@@ -118,6 +119,14 @@ function parseMarkdownContent(data) {
 	});
 }
 
+function addFragmentLinksToHeadings(data) {
+	data.html = data.html.replaceAll(
+		/(<h[1-6] id="([^"]+)">)([^\n]+)/g,
+		'$1<a href="#$2"></a>$3'
+	);
+	return data;
+}
```
marvinhagemeister, line 128: If we upgrade `marked` we can hook directly in their markdown renderer instead of regexing the HTML after all is done.

rschristian: Oh neat, didn't realize it supported that. Will take a look tomorrow

rschristian: Oh man, 9 major versions. Will need time to verify everything's working correctly

rschristian: So `marked` has tore through 5 majors in the past 4 months, including ripping some things we use. I don't think it makes sense to touch that now, especially as the docs don't seem to be in a good state.

developit:
This is a good idea. I agree with avoiding the Marked upgrade, that sounds way too painful.

As an experiment, we could use the HTML-element-to-component mapping layer we have in place to remap `h[1-6]` to components that render the heading and link as a pair:
https://github.com/preactjs/preact-www/pull/1013

marvinhagemeister (approved):

PR #1013, opened 2023-09-14
heading links alt

developit:
an alternative approach to #1012 for consideration


rschristian:
Ah, didn't realize we already had some element manipulation going on, must've missed this with my greps. Yeah would make more sense to do it this way

rschristian:
Merged into #1012

PR #1014, opened 2023-09-29
Fix signals REPL example

### LINKED ISSUE #978: Signals example doesn't work
https://preactjs.com/repl?example=todo-list-signals

Expected behavior: "a fully working todo app" as the link on https://preactjs.com/guide/v10/signals#building-the-ui calls it.

Observed behavior: page loads, but no functionality (add, check, delete all do nothing)

I tried Chrome and Safari on my Mac, with extensions disabled, so I think I've ruled out that it's a local browser issue.

andrewiggins:
Signals REPL example was failing cuz we were loading two copies of Preact on the page. The repl file imported 'preact' which esm.sh resolved to `esm.sh/preact@10.18.0`. The repl file also imported `@preact/signals` which also imports `preact` but esm.sh puts the version from the signals package.json on that import so the signals package imported `preact@10.17.0`.

This PR fixes the REPL by forcing all esm.sh URLs to have the `external=*` query param meaning any external imports are imported using bare specifiers (e.g. `preact` instead of `preact@10.17.0`). This ensures any imports transitive imports from esm.sh go through our bundler to get resolved, and we can resolve them to the same URL.

I've also updated the rollup config to use `resolveId` to resolve bare module specifiers to esm.sh URLs simplifying the logic a bit I think.

This does mean that if you import a library that breaks with the latest version of a package, it won't work in the REPL :/ But I figured getting something working for our examples and for Preact libs would be a good start.

Resolves #978

### DIFF
--- a/src/components/controllers/repl/examples/todo-list-signal.txt
+++ b/src/components/controllers/repl/examples/todo-list-signal.txt
@@ -34,18 +34,18 @@ function TodoList() {
         {todos.value.map((todo, index) => {
           return (
             <li>
-							<label>
-								<input
-									type="checkbox"
-									checked={todo.completed}
-									onInput={() => {
-										todo.completed = !todo.completed
-										todos.value = [...todos.value];
-									}}
-								/>
-								{todo.completed ? <s>{todo.text}</s> : todo.text}
-							</label>
-							{' '}
+              <label>
+                <input
+                  type="checkbox"
+                  checked={todo.completed}
+                  onInput={() => {
+                    todo.completed = !todo.completed
+                    todos.value = [...todos.value];
+                  }}
+                />
+                {todo.completed ? <s>{todo.text}</s> : todo.text}
+              </label>
+              {' '}
               <button onClick={() => removeTodo(index)}>❌</button>
             </li>
           );
--- a/src/components/controllers/repl/examples/todo-list-signal.txt
+++ b/src/components/controllers/repl/examples/todo-list-signal.txt
@@ -34,15 +34,18 @@ function TodoList() {
         {todos.value.map((todo, index) => {
           return (
             <li>
-              <input
-                type="checkbox"
-                checked={todo.completed}
-                onInput={() => {
-                  todo.completed = !todo.completed
-                  todos.value = [...todos.value];
-                }}
-              />
-              {todo.completed ? <s>{todo.text}</s> : todo.text}{' '}
+							<label>
+								<input
+									type="checkbox"
+									checked={todo.completed}
+									onInput={() => {
+										todo.completed = !todo.completed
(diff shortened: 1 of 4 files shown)


src/components/controllers/repl/examples/todo-list-signal.txt
```diff
@@ -34,15 +34,18 @@ function TodoList() {
...
+									type="checkbox"
+									checked={todo.completed}
+									onInput={() => {
+										todo.completed = !todo.completed
+										todos.value = [...todos.value];
+									}}
+								/>
+								{todo.completed ? <s>{todo.text}</s> : todo.text}
+							</label>
+							{' '}
```
rschristian, line 48: nit: formatting is a bit off

```suggestion
              <label>
                <input
                  type="checkbox"
                  checked={todo.completed}
                  onInput={() => {
                    todo.completed = !todo.completed
                    todos.value = [...todos.value];
                  }}
                />
                {todo.completed ? <s>{todo.text}</s> : todo.text}
              </label>
              {' '}
```

rschristian (approved):
It'll be great to have this working again!

src/components/controllers/repl/repl.worker.js
```diff
@@ -84,30 +84,48 @@ async function bundle(sources) {
...
 						const code = sources[id];
 						const out = transpile(code);
 						return out;
 					}
-					if (id.startsWith('https://esm.sh/')) return get(id);
-					const [, mod, v, path] =
-						id.match(/^((?:@[^@/]+\/)?[^@/]+)(?:@(\d[^@/]*))?(?:\/(.*))?$/) ||
-						[];
-					return dep(mod, v, path);
+					return get(id);
```
andrewiggins, line 128: Wasn't sure what this doing but I've removed it now since we resolve all module specifiers to esm.sh URLs in `resolveId()` instead of `load()` now. As its written now, you can specify any syntax or options esm.sh supports (e.g. `import useSWR from "swr@^2?alias=react:preact/compat";`)

andrewiggins: Okay this particular example doesn't work cuz setting `external` breaks downstream aliasing but removing `external` doesn't work either cuz `https://esm.sh/swr@2.2.4?alias=react:preact/compat` will attempt to import `/stable/preact@10.17.1/es2022/compat.js` instead of the latest version 🤦‍♂️. Since we just published a new version, perhaps this is just some temporary version caching delay?

andrewiggins: Okay aliasing just doesn't work currently. If I manually specifiy 10.17.1 everywhere I can, some transitive dependency will still pull in 10.18.0 lol

PR #1015, opened 2023-09-29
refactor: Document error boundaries specifically

rschristian:
Inspired by https://x.com/windchime_yk/status/1707583523052453934?s=20

Mostly just to make error boundary info more discoverable, whether you're coming from a search engine or just searching our own docs, as we never actually use "error boundary" despite that being the prevailing term.

### DIFF
--- a/content/en/guide/v10/components.md
+++ b/content/en/guide/v10/components.md
@@ -96,25 +96,28 @@ In order to have the clock's time update every second, we need to know when `<Cl
 
 > See [this diagram](https://twitter.com/dan_abramov/status/981712092611989509) to get a visual overview of how they relate to each other.
 
-#### componentDidCatch
+### Error Boundaries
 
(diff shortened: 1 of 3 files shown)


content/en/guide/v10/components.md
```diff
@@ -96,25 +96,28 @@ In order to have the clock's time update every second, we need to know when `<Cl
...
   }
 
-  componentDidCatch(error) {
-    this.setState({ errored: true });
+  static getDerivedStateFromError(error) {
+    return { errored: true };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    errorReportingService(error, errorInfo);
```
rschristian, line 120: React deprecated setting state from `componentDidCatch` at some point but I'm not sure if we care about that at all? Error reporting seems to be its intended purpose at this point: https://react.dev/reference/react/Component#componentdidcatch

JoviDeCroock (approved):

PR #1026, opened 2023-11-27
Add Russian translation

dragomano:


rschristian (approved):
Wow, this is huge! Thanks!

Just to check, as I imagine this amount of changes means you fed it into a translator, were you able to give it a good skim to ensure everything looks reasonable?

dragomano:
Don't worry. I tried to view each page in the VSCode preview tool and also tested the code in practice. After posting on the site, I will slowly check and correct inaccuracies if I find any.

rschristian:
Thanks, just wanted to confirm.

PR #822, opened 2022-02-07
Add todo item removal to todo-list repl example

zadeviggers:

PR #1027, opened 2023-11-27
refactor: Flesh out 'Todo List' REPL example to match 'Todo List (Signals)'

rschristian:
Someone mentioned this ages ago and it's been on my todo list (hehe) since.

If we have `Todo List` and `Todo List (Signals)`, it should be expected that the functionality otherwise matches and is more useful at comparing the class component to signals implementations. I tried to keep them looking pretty similar where possible, even if this might be a bit more verbose than you'd usually see.

Supersedes #822


rschristian:
Superseded by #1164

PR #1032, opened 2023-12-03
refactor: Merge header & sponsor SVGs into sprite sheets

rschristian:
Few less requests (9 -> 3), bit smaller transfer size (68.29kb -> 61.76kb), no change is usability.

I did take a look at merging the other SVGs on the home page, but we actually benefit from the lazy loading there a fair bit -- on most screens the first icon will immediately loaded, which would drag in the entire sprite sheet. This is a ~6.5kb difference on page load, so not worth it.


src/components/header/index.js
```diff
@@ -46,24 +46,18 @@ export default function Header() {
 							aria-label="Browse the code on GitHub"
 							href="https://github.com/preactjs/preact"
 						>
-							<img
-								src="/assets/github.svg"
-								alt="GitHub"
-								width="34"
-								height="33"
-							/>
+							<svg aria-hidden viewBox="0 0 24 24">
```
rschristian, line 49: _I think_ we're meant to use aria-hidden here, as it's just content of the link and not relevant to screen readers?

src/components/sponsors/style.module.less
```diff
@@ -34,27 +34,32 @@
...
 					drop-shadow(0 0 20px rgba(255, 255, 255, 0.5));
 			}
 		}
 	}
 
-	img {
-		height: 4.5rem;
+	img,
+	svg {
+		height: 3rem;
```
rschristian, line 61: We had inline styles setting `height: 3rem`, so moved here.

JoviDeCroock (approved):

PR #1035, opened 2023-12-04
chore: Add scrollbar-color to styles

lukewarlow:
This adds `scrollbar-color` styles so that the look and feel of scrollbars on Firefox matches Chrome and Safari.

This property is supported in Firefox and soon to be Chrome and overrides the `::-webkit-scrollbar` styles in Chrome.


rschristian:
Thanks!

PR #1036, opened 2023-12-05
refactor: Switch off LESS to vanilla CSS

rschristian:
Easier to scan the diff on GitHub while I work than locally 😅

Toying with switching to plain CSS -- it's gotten pretty darn good and we don't make use of that much of LESS in all reality (15 vars & 5 mixins total, and some of these are long unused). There's a couple areas that get slightly more fiddly (the `sidebar-break` variable in particular -- we use it for media queries but that's still not supported in standard CSS, sadly) but by and large things are unchanged.

TODO:
  - ~~Enable PostCSS nesting plugin for old browsers, probably don't want to ship nested CSS~~ **DONE**
  - Check how well `filter: brightness()` is supported for swapping out `lighten()` and `darken()` utilities. Right now I'm just grabbing the computed hex value from the browser and reinserting with a comment
    - Can be done later, if we wish. Comments exist stating where the color came from


marvinhagemeister:
+1 I like the move to vanilla css. Grabbing the actual values from `lighten()` etc seems fine 👍

src/style/reset.less
```diff
@@ -1,8 +0,0 @@
-ol,
-ul {
-	margin-left: 0;
-}
-
-select {
-	font-size: 100%;
-}
```
rschristian, line 8: Moved into `index.css`

src/components/branding/style.module.css
```diff
@@ -1,5 +1,3 @@
-@import '~style/helpers';
```
rschristian, line 1: Interestingly, this change shaves a huge chunk off our CSS bundles -- every import of `helpers.less` here (which in turn imported the `variables.less`) resulting in all of our `:root { }` CSS vars ending up in the bundle.

Loading the tutorial, for example, loads the page with 4 copies of our CSS vars. Yikes.

rschristian: Wow, reduces transfer size by 4.5kb and total size by nearly 7kb.

The page is still 381kb transferred so it's not a measurable improvement of course, just nice to see.

src/components/controllers/page/sidebar.module.less
```diff
@@ -44,12 +44,12 @@
...
 		// Use box shadow to draw an outline because outline has no radius
-		box-shadow: 0 0 2px 2px lighten(@color-brand, 40%),
-			0 3px 8px rgba(0, 0, 0, 0.5);
+		box-shadow: 0 0 2px 2px #dacef0,
+			/* --color-brand 40% lighter */ 0 3px 8px rgba(0, 0, 0, 0.5);
 	}
 
 	[data-open] & {
-		background: @color-corner;
+		background: var(--color-brand-light);
```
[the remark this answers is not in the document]
marvinhagemeister: I think it was intentional on my part. Don't mind changing it though

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #1041, opened 2023-12-20
fix: blog page title invisible issue

n0hack:
## What is this PR?

I have fixed the issue of the post titles being invisible on the blog page. It was found to be influenced by the parent styles, and I have temporarily addressed it using `!important`.

### Before

image

### After

image

### DIFF
--- a/src/components/blog-overview/blog.module.less
+++ b/src/components/blog-overview/blog.module.less
@@ -10,6 +10,8 @@
 	margin-top: 0 !important;
 
 	a {
+		width: 100% !important;
+		opacity: 1 !important;
 		color: var(--color-link);
 	}
 }
--- a/src/config.json
+++ b/src/config.json
@@ -51,6 +51,7 @@
 			"de": "Weiter lesen",
 			"zh": "继续阅读",
 			"es": "Seguir leyendo",
+			"kr": "계속 읽기",
 			"ru": "Продолжить чтение"
 		},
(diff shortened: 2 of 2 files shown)


rschristian:
Ah shoot, I must've broken in with #1012. Probably need to limit those styles entirely to `content-region[name*='guide']`

n0hack:
@rschristian then enclosing only this part with `content-region[name*='guide']` should be sufficient?

**markdown.less**
```css
content-region[name*='guide'] {
	a {
		opacity: 0;
		display: inline-flex;
		width: 1.5rem;
		margin-left: -1.5rem;
		&::before {
			content: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="%23fff" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M15 7h3a5 5 0 0 1 5 5 5 5 0 0 1-5 5h-3m-6 0H6a5 5 0 0 1-5-5 5 5 0 0 1 5-5h3"></path><line x1="8" y1="12" x2="16" y2="12"></line></svg>');
		}
	}
}
```

rschristian:
Something along those lines I think, yeah. Not 100% sure off the top of my head, would need to test it

n0hack:
I'll go ahead with the translation work first, and I'll test this later

PR #1045, opened 2023-12-31
Add Spanish translation of Forms and Hooks

Nicolas-Orozco:


marvinhagemeister (approved):
Sweet, thanks for adding this 👍

PR #1051, opened 2024-01-13
Add documentation for clean up functions in effects

gabrielcoronel:
The current documentation did not mention the possibility of specifying clean up functions for effects at all, which might be confusing since such possibility is only obvious from reading the source code.

These remarks were added in the effects section, together with a code example, and its corresponding API section.

### DIFF
--- a/content/en/guide/v10/signals.md
+++ b/content/en/guide/v10/signals.md
@@ -356,7 +356,7 @@ name.value = "John";
 // Logs: "John Doe"
 ```
 
-You can return a function from the function provided to [`effect()`](#effectfn) that will be run right before the next update takes place. The purpose of the returned function is to clean up the side effect before the next update.
+Optionally, you can return a cleanup function from the callback provided to [`effect()`](#effectfn) that will be run before the next update takes place. This allows you to "clean up" the side effect and potentially reset any state for the subsequent trigger of the callback.
 
 ```js
 effect(() => {
--- a/content/en/guide/v10/signals.md
+++ b/content/en/guide/v10/signals.md
@@ -356,6 +356,16 @@ name.value = "John";
 // Logs: "John Doe"
 ```
 
+You can return a function from the function provided to [`effect()`](#effectfn) that will be run right before the next update takes place. The purpose of the returned function is to clean up the side effect before the next update.
+
+```js
+effect(() => {
+  Chat.connect(username.value)
+
+  return () => Chat.disconnect(username.value)
+})
+```
+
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/signals.md
```diff
@@ -356,6 +356,16 @@ name.value = "John";
 // Logs: "John Doe"
 ```
 
+You can return a function from the function provided to [`effect()`](#effectfn) that will be run right before the next update takes place. The purpose of the returned function is to clean up the side effect before the next update.
```
rschristian, line 359:
```suggestion
Optionally, you can return a cleanup function from the callback provided to [`effect()`](#effectfn) that will be run before the next update takes place. This allows you to "clean up" the side effect and potentially reset any state for the subsequent trigger of the callback.
```

I don't think this section fits where you've inserted it though, would rather have this example & description in the API docs below.

gabrielcoronel: As far as I understand the API docs are supposed to be reference for people who already know how a feature works. So people should be able to see a short explanation and example of cleanup functions before going to the reference.

content/en/guide/v10/signals.md
```diff
@@ -506,7 +516,7 @@ When creating computed signals within a component, use the hook variant: `useCom
 
 ### effect(fn)
 
-To run arbitrary code in response to signal changes, we can use `effect(fn)`. Similar to computed signals, effects track which signals are accessed and re-run their callback when those signals change. Unlike computed signals, `effect()` does not return a signal - it's the end of a sequence of changes.
+To run arbitrary code in response to signal changes, we can use `effect(fn)`. Similar to computed signals, effects track which signals are accessed and re-run their callback when those signals change. If `fn` returns a function, this function will be run right before the next value update. Unlike computed signals, `effect()` does not return a signal - it's the end of a sequence of changes.
```
rschristian, line 519:
```suggestion
To run arbitrary code in response to signal changes, we can use `effect(fn)`. Similar to computed signals, effects track which signals are accessed and re-run their callback when those signals change. If the callback returns a function, this function will be run before the next value update. Unlike computed signals, `effect()` does not return a signal - it's the end of a sequence of changes.
```

rschristian (approved):
Cheers!

content/en/guide/v10/signals.md
rschristian: I don't really agree there, API docs are for both new users and familiar ones alike, but we can keep this I suppose.

PR #1052, opened 2024-01-16
chore: Switch to `netlify-cli`

rschristian:
I don't really intend to merge this but wanted to have it in case we are forced to switch off `netlify-lambda` (which has been deprecated and relies on Webpack 4, which has issues past Node 17).

`netlify-cli` is absurdly large (as you can tell by the diff), adds ~1300 deps and increases `node_modules` from 481M to 784M. IMO, this is a complete non-starter. Almost certainly better off going with a custom build/serve set up.


rschristian:
Closing in favor of #1108, we shouldn't need this anymore.

PR #1053, opened 2024-01-18
refactor: Switch to Vite

rschristian:
I tried to keep this somewhat contained, but it's enormous. Sorry!


patches/netlify-lambda+1.6.3.patch
```diff
@@ -0,0 +1,35 @@
+diff --git a/node_modules/netlify-lambda/lib/build.js b/node_modules/netlify-lambda/lib/build.js
+index ac59043..6a07f3f 100644
+--- a/node_modules/netlify-lambda/lib/build.js
++++ b/node_modules/netlify-lambda/lib/build.js
+@@ -109,7 +109,7 @@ function webpackConfig(dir, { userWebpackConfig, useBabelrc } = {}) {
+     ],
+     output: {
+       path: functionsPath,
+-      filename: "[name].js",
++      filename: "[name].cjs",
```
rschristian, line 10: We can likely get away with getting rid of `netlify-lambda` entirely -- it's deprecated but the replacement, `netlify-cli`, ain't great. See #1052 

All we need is to compile the functions and, for local use, serve them.

src/components/code-editor/index.jsx
```diff
@@ -1,10 +1,12 @@
-import { h, Component, render } from 'preact';
+import { Component, render } from 'preact';
 import codemirror from 'codemirror';
 import cx from '../../lib/cx';
 import 'codemirror/mode/jsx/jsx';
 import 'codemirror/addon/comment/comment';
 import 'codemirror/lib/codemirror.css';
 import style from './style.module.css';
+// TODO: Fix me, Vite doesn't seem to like :global in CSS Modules
+import './temp.css';
```
rschristian, line 9: Reminder for myself

rschristian: More specifically, Vite's CSS Modules processor (whatever it is) dislikes `:global {}` blocks.

This is fine, larger blocks we can just extract out to separate plain `.css` files.

src/components/controllers/style.module.css
```diff
@@ -34,7 +34,7 @@
 			padding-left: 50px;
 		}
 
-		.full-width {
+		:global(.full-width) {
```
rschristian, line 37: Worth calling out that Vite, or whatever Vite's using to process CSS Modules, seemingly has a rather different interpretation of `:global` than other tools. We should probably try to get rid of all usages of `:global` in the future, as it's pretty fiddly.

vite.config.js
```diff
@@ -0,0 +1,162 @@
...
+		content = '---\n' + JSON.stringify(meta) + '\n---\n' + content;
+	}
+
+	return content;
+}
+
+export default defineConfig({
+	publicDir: 'src/assets',
+	build: {
+		target: ['chrome88', 'edge88', 'es2020', 'firefox78', 'safari14'],
```
rschristian, line 49: Default is this list but `chrome87`, not 88.

Unfortunately Vite's CSS transpiler cannot handle a few of our nested classes without generating an `is:(...)`, which is only available in chrome88. The tiny bump from defaults felt warranted, though we can discuss other targets. Might need some refactoring is all.

patches/preact-iso+2.2.0.patch
```diff
@@ -0,0 +1,14 @@
...
+index aaeba33..538096c 100644
+--- a/node_modules/preact-iso/prerender.js
++++ b/node_modules/preact-iso/prerender.js
+@@ -46,7 +46,8 @@ export default async function prerender(vnode, options) {
+ 	};
+ 
+ 	try {
+-		const html = await render();
++		let html = await render();
++		html += `<script type="isodata"></script>`;
```
rschristian, line 11: Backporting the `isodata` script -- our routing here has some problems on 2.3.0 that I haven't been able to address, so patching was easier.

patches/vite-plugin-static-copy+1.0.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/vite-plugin-static-copy/dist/index.js b/node_modules/vite-plugin-static-copy/dist/index.js
+index 3c00baa..e94a264 100644
+--- a/node_modules/vite-plugin-static-copy/dist/index.js
++++ b/node_modules/vite-plugin-static-copy/dist/index.js
+@@ -1064,7 +1064,7 @@ var buildPlugin = ({
+     buildEnd() {
+       output = false;
+     },
+-    async writeBundle() {
++    async generateBundle() {
```
rschristian, line 10: Need to move this up for prerendering.

patches/netlify-lambda+2.0.16.patch
```diff
@@ -0,0 +1,35 @@
+diff --git a/node_modules/netlify-lambda/lib/build.js b/node_modules/netlify-lambda/lib/build.js
+index 4397356..e07c889 100644
```
rschristian, line 2: We should be able to get rid of `netlify-lambda` in general, swapping it out for a middleware in dev and shipping raw functions in prod, but that can be addressed later.

src/pwa.js
```diff
@@ -1,38 +0,0 @@
-let hasInteracted, shouldReload;
```
rschristian, line 1: It's possible I'm missing something, but I don't see much of a reason to keep this as a PWA, at least now without altering our set up quite a bit. I don't think we're getting much out of it being a PWA at the moment.

marvinhagemeister: +1 on dropping PWA mode. It was a cool tech demo, but I'm skeptical that our PWA is even used these days.

marvinhagemeister (approved):

rschristian:
'Ron Paul - It's happening' gif

PR #1054, opened 2024-01-20
fix: Browser page titles

rschristian:
This addresses a few things:

- REPL doesn't set the page (browser) title
  - This can be seen most easily by opening [this link](https://preactjs.com/guide/v10/getting-started) to the 'Getting Started' page and then switching to the REPL from the header -- the browser title will still be 'Getting Started ...'
- The prerendered titles do not match the titles as they'd be set client-side
  - I like the prerendered titles way more, `Getting Started - Preact Guide` vs `Getting Started | Preact: Fast 3kb React alternative with the same ES6 API. Components & Virtual DOM.` See below, I think this might've been a bug.
- Upon extracting out the title creation, I also tried to make `/page` less of a grab-bag and moved most of the main utilities out. It's still not ideal and can use some refactoring (splitting it up into guide, tutorial, and blog would probably be a good idea), but this is an improvement I think.
- Slightly edited the `showTitle` for Page, see below for explanation

### DIFF
--- a/src/components/controllers/page/index.js
+++ b/src/components/controllers/page/index.js
@@ -1,131 +1,19 @@
 import { Fragment, h } from 'preact';
-import { useEffect, useState, useMemo, useRef } from 'preact/hooks';
+import { useEffect, useMemo } from 'preact/hooks';
 import cx from '../../../lib/cx';
 import ContentRegion from '../../content-region';
-import { getContentOnServer, getContent } from '../../../lib/content';
 import config from '../../../config.json';
 import style from './style.module.css';
 import Footer from '../../footer';
 import Sidebar from './sidebar';
 import Hydrator from '../../../lib/hydrator';
 import EditThisPage from '../../edit-button';
-import {
-	getPrerenderData,
-	InjectPrerenderData
-} from '../../../lib/prerender-data';
+import { InjectPrerenderData } from '../../../lib/prerender-data';
 import { isDocPage } from '../../../lib/docs';
 import { useStore } from '../../store-adapter';
 import { AVAILABLE_DOCS } from '../../doc-version';
 import { Time } from '../../time';
-
-const getContentId = route => route.content || route.path;
-
-/**
- * Set `document.title`
- * @param {string} title
- */
-export function useTitle(title) {
-	useEffect(() => {
-		if (title) {
-			document.title = `${title} | ${config.title}`;
-		}
-	}, [title]);
-}
-
-/**
- * Set the meta description tag content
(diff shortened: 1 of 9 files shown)


src/components/controllers/page/index.js
```diff
@@ -1,131 +1,19 @@
...
-const getContentId = route => route.content || route.path;
-
-/**
- * Set `document.title`
- * @param {string} title
- */
-export function useTitle(title) {
-	useEffect(() => {
-		if (title) {
-			document.title = `${title} | ${config.title}`;
```
rschristian, line 30: I feel like this was meant to be an `||`? `Getting Started | Preact: Fast 3kb React alternative with the same ES6 API. Components & Virtual DOM.` is quite a bit of text for a title, but it could make sense as a fallback?

src/components/controllers/page/index.js
```diff
@@ -157,8 +45,9 @@ export default function Page({ route, prev, next }, ctx) {
 	// "name" is the exact page ID from the URL
 	// "current" is the currently *displayed* page ID.
 
-	const showTitle = current != 'index' && meta.show_title !== false;
-	const canEdit = showTitle && current != '404' && current !== '/blog';
+	const isHome = current === 'index';
+	const showTitle = meta.show_title !== false;
+	const canEdit = !isHome;
```
rschristian, line 50: `showTitle`, as far as I can tell, is only used to differentiate the home page from the rest of the content. The problem I had with this is that I thought the spacing on every other page was rather nice, as it kept the content from bunching up against the header on pages where we didn't have a title, such as the blog:

With Spacer             |  Without Spacer
With spacer |  Without spacer

As such, I split this up into `isHome` and `showTitle`. As for `canEdit`, I tend to that's fine to show on every page besides home? Happy to revert though.

Coincidentally, on `/blog` the title wasn't supposed to be shown at all but `current` is actually `blog`, not `/blog`. I kinda prefer it though?

rschristian:
Whoops think I broke something on the homepage.

marvinhagemeister (approved):
Nice, those cleanups are great! I'm neutral on whether there is an edit button on the blog posts or not. It's more important to have it for the docs pages. Really nice work as usual!

rschristian:
Only just remembered the edit button doubles as the "add a translation" when one's missing which I think might be a better case for it to stay. Someone's (kindly) translated the blog posts into Korean, but not the blog page itself, so I worry it might be a bit jarring without a warning? Not sure. I could go either way.

https://preactjs.com/blog/?lang=kr

PR #1055, opened 2024-01-20
fix: Make footer sticky

rschristian:
I have a taller browser than most (1440p), but this can also be seen on mobile on a few pages.

Before            |  After
Before PR | After PR


marvinhagemeister (approved):
Good catch!

PR #1057, opened 2024-01-21
fix: Correct background colors on tutorial & guide pages

### LINKED ISSUE #1056: Page background changes color depending on sidebar
Reminder for myself:

[Getting Started](https://preactjs.com/guide/v10/getting-started/): #1a1a1a
[Blog](https://preactjs.com/blog): #242424

https://github.com/preactjs/preact-www/blob/2f102c94f053e860a27bef295e970cf8ae67dd6d/src/components/controllers/page/style.module.css#L117

Darker looks better & has higher contrast. Should just switch to it everywhere.

Edit: Hm, darker does result in the footer blending in quite a bit though.

rschristian:
Closes #1056 

Most obviously seen by going to the [the blog](https://preactjs.com/blog) with a dark mode preference and then navigating to the guide via the header. Light mode also has a difference but it's much smaller and therefore not as obvious.

I'm guessing this was a bug from building out the sidebar on mobile, accidentally always applying the background color to the guide. Tutorial simply copied this coloring.

That being said, either color looks great, could go either way with it.


src/components/controllers/tutorial/style.module.css
```diff
@@ -393,10 +390,5 @@
 }
 
 .output {
-	color: #444;
 	height: 100%;
-
-	h1 {
-		color: #444;
-	}
```
rschristian, line 401: Should be entirely unused, as we load the iframe as a child of `.output` here. Spotted when playing around with the colors.

marvinhagemeister (approved):

PR #1058, opened 2024-01-21
refactor: Reduce content tearing

rschristian:
This addresses a couple different tearing issues:

- When switching to larger pages, you can sometimes spot the title and content tearing due to the title being extracted and rendered separately from the rest of the content. This is occasional, usually only on larger pages, but this set up feels a fair bit simpler anyhow? Please correct me if I'm missing context, I couldn't really figure out why the title was extracted out in the first place by looking through the blame (besides maybe styling? but that's an easy fix).
- When switching from the home page to one with an "Edit this page" button, that would usually appear considerably before the home page's content got swapped out, due to `content-region` using the route name (which updates immediately) rather than the current shown content's name. Simple fix, switch to current content name to keep the edit button in sync with page content.
  - This is really bad at the moment due to my changes in #1054, didn't catch it. Going from the home page to say the blog shows a black bar underneath the header **really** early.

If you're comparing this PR to live, there's a slight reduction in spacing below the header (regardless of whether there's a page title or not) as the blog meta data (authors & translated by) was always present and creating a bit of padding, even if it had no text content outside of blog articles. The home page's spacing is also slightly off due to it having 4 `<h1>`'s, which it probably shouldn't (will fix separately soon-ish). Both are very minor shifts.

Content diff is a bit busy, but all that was done was adding headers in the markdown content for the 'We Are Using' page and blog posts, and removing the now unused `show_title` value in the frontmatter.


src/components/controllers/page/style.module.css
```diff
@@ -7,6 +7,14 @@
 		margin-bottom: 4rem;
 	}
 
+	content-region:not([name='index']) :global(.markup) {
+		margin-top: 2.5rem;
```
rschristian, line 11: TODO: This unfortunately causes a black bar to appear beneath the header before the rest of the page switches -- `content-region` updates near immediately, while the actual content takes a bit to load. Will need to rethink.

This already exists somewhat but the change here exacerbates it.

marvinhagemeister (approved):

PR #1059, opened 2024-01-22
refactor: Ensure landing page only has a single `<h1>`

rschristian:
Whether or not multiple `<h1>` are "valid" is quite murky at the moment, as while the WHATWG HTML spec says it's a-ok if they're in `<section>`s, reality is that a lot of browsers and accessibility tools don't quite follow or acknowledge the nesting "as they should" (with some arguments being that the spec is problematic and should not be followed). This has been the case for well over a decade now and is mostly in a stand-still from my understanding.

It's easy enough for us to "correct" anyhow, a single `<h1>` for the entire page followed by `<h2>` for the rest conveys the meaning we need.

### DIFF
--- a/src/style/index.css
+++ b/src/style/index.css
@@ -47,7 +47,11 @@ a {
 }
 
 h1,
-h2 {
+h2,
+h3,
+h4,
+h5,
+h6 {
 	color: var(--color-heading);
 }
 
--- a/content/de/index.md
+++ b/content/de/index.md
@@ -37,7 +37,7 @@ function Counter() {
 </div>
 
 <section class="home-top">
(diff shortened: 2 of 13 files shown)


src/style/index.css
```diff
@@ -47,7 +47,11 @@ a {
 }
 
 h1,
-h2 {
+h2,
+h3,
+h4,
+h5,
+h6 {
```
rschristian, line 54: IMO, looks better applying this to all headings. I think the deepest nesting we have is our ["Aliasing in X" instructions](https://preactjs.com/guide/v10/getting-started#aliasing-in-webpack) and the `<h4>` titles really don't stand out too well. This helps.

PR #1061, opened 2024-01-23
chore: Bump dev deps, removed unused deps, correct Netlify redirects, add `jsconfig.json`

rschristian:
Few less warnings upon install and signals of outdated deps in the `package.json`

Some of the bigger runtime deps seem to have issues (bumping Preact does some weird stuff to the header in particular), w/ look into those separately.


jsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
rschristian, line 1: There's a, uh, fair number of red squiggles now, but most of them are things we should probably be tracking.

Dunno about other editors, TS only runs in mine w/ a `js/tsconfig` available

preact.config.js
```diff
@@ -114,14 +112,14 @@ export default function (config, env, helpers) {
...
+				context: __dirname,
+				from: 'src/_redirects',
+				to: '_redirects',
 				toType: 'file'
 			}]
 		);
-
-		netlifyPlugin(config, {
-			redirects: fs.readFileSync('src/_redirects', 'utf-8').trim().split('\n')
-		});
```
rschristian, line 124: Accidentally broken with the v4 migration, simply checked out an older version of the repo to see what the plugin added and it was only a single redirect. Easy enough to just copy into source.

src/lib/router.js
```diff
@@ -1,46 +0,0 @@
-import { useEffect, useState } from 'preact/hooks';
```
rschristian, line 1: Looks to have been unused for years.

package.json
```diff
@@ -1,25 +1,22 @@
...
+    "node": ">=18"
   },
   "scripts": {
     "dev": "concurrently \"npm run dev:client\" \"npm run dev:lambda\"",
     "dev:client": "preact watch --no-refresh",
     "dev:lambda": "netlify-lambda serve src/lambda",
     "start": "npm run -s server",
     "prestart": "npm run -s build",
-    "server": "superstatic build -p ${PORT:-8080} --host 0.0.0.0 --gzip -c '{\"rewrites\": [{\"source\":\"**\",\"destination\":\"index.html\"}],\"headers\":[{\"source\":\"**\",\"headers\":[{\"key\":\"Cache-Control\",\"value\":\"max-age=31536000\"}]}]}'",
+    "server": "superstatic build -p ${PORT:-8080} --host 0.0.0.0 --compression -c '{\"headers\":[{\"source\":\"**\",\"headers\":[{\"key\":\"Cache-Control\",\"value\":\"max-age=31536000\"}]}]}'",
```
rschristian, line 14: Doesn't look like the rewrite works but we shouldn't want that as we are prerendering all pages.

rschristian, line 14: AFAICT, the rewriting doesn't work and I don't think we'd want that? Ideally we'd have a 404 fallback but I can't see how to easily set that up.

marvinhagemeister (approved):

PR #1064, opened 2024-01-26
docs: TS switch from `type` to `interfaces`

marvinhagemeister:
These days the TypeScript recommends to use the `interface` keyword instead of `type` in their docs:

https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html

> You’ll see that there are two syntaxes for building types: Interfaces and Types. You should prefer interface. Use type when you need specific features.

For typing component props whether you use `type` or `interface` hardly matters, since the differences mainly show up when you combine or extend types which is rare with props. That said it's better for us to align with the TS docs.

### DIFF
--- a/content/en/guide/v10/typescript.md
+++ b/content/en/guide/v10/typescript.md
@@ -99,7 +99,7 @@ There are different ways to type components in Preact. Class components have gen
 Typing regular function components is as easy as adding type information to the function arguments.
 
 ```tsx
-type MyComponentProps = {
+interface MyComponentProps {
   name: string;
   age: number;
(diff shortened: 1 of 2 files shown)


rschristian (approved):
Should probably switch all translations over but I can do that later if it's a hassle

marvinhagemeister:
Good point about the other languages. Did the PR from the web editor and didn't think of that. Updated the other languages too.

rschristian:
Ideally we'd have a better system to keep code blocks in sync, but... something for future us to think about 😅

PR #1065, opened 2024-01-26
docs: update zh getting-started

rxliuli:
Currently, the **Getting Started > Create a Vite-Powered Preact App** section is not translated into Chinese, and it still recommends using the old **Preact CLI**. This inconsistency on the first page is quite misleading, so I making a modification.


rschristian (approved):
Thanks!

Languages other than English are entirely community-contributed, we don't maintain the content as none of us speak the language.

PR #1069, opened 2024-02-04
refactor: Keep tutorial titles in w/ markdown content

rschristian:
Similar to #1058, but for the tutorial pages.

No real need to extract it and keeping it together ensures less tearing.


src/components/controllers/tutorial/index.js
```diff
@@ -278,8 +277,6 @@ function TutorialView({
 				}
 			>
 				<div class={style.tutorialWindow} ref={content}>
-					<h1 class={style.title}>{title}</h1>
```
rschristian, line 281: `style.title` doesn't actually exist so it didn't need to be cleaned up in the CSS.

marvinhagemeister (approved):

PR #1073, opened 2024-02-05
chore: Backport much of #793

rschristian:
Vast majority of this is Marvin's. Had some great stuff sitting around in the WMR PR that I wanted to nab

Does 3 major things:

1. Switches to `preact-iso` for routing, lazy loading.
2. Gets off `unistore` to `Context` for most items
3. Breaks up the reliance on `controllers/page` and creates separate DocPage, Blog, etc. Much cleaner, much simpler.
    - Tutorial & REPL could use a similar treatment but they'd be a bigger refactor

I tried to keep the diff down but... this is gonna be a big one. Sorry.

Shouldn't be any functional differences on the site.


package.json
```diff
@@ -88,12 +90,11 @@
     "node-fetch": "^2.6.1",
     "preact": "10.15.1",
     "preact-custom-element": "^4.3.0",
+    "preact-iso": "2.2.0",
```
rschristian, line 93: Something in 2.3.0+ (I'm guessing https://github.com/preactjs/wmr/pull/864) results in `onLoadEnd` never being called on first page rendered, resulting in our loading bar going until the user switches pages.

src/components/code-block/prism.worker.js
```diff
@@ -7,7 +7,8 @@ export function highlight(code, lang) {
 	if (prism.languages[lang] != null) {
 		return prism.highlight(code, prism.languages[lang], lang);
 	}
-	throw Error(`Unknown language: ${lang}`);
+	//console.error(`Unknown language: ${lang}`);
+	return code;
```
rschristian, line 11: We're missing `bash` & `diff`, both are rather inconsequential though.

src/components/controllers/tutorial/index.js
```diff
@@ -158,44 +180,52 @@ export default class Tutorial extends Component {
...
-			getContent([lang, page.meta.next]);
-		}
-	}, [page.meta && page.meta.next, fullPath]);
+	// TODO: Webpack creates a circular dependency that
+	// it cannot resolve. Temporarily disabled
+	//useEffect(() => {
+	//	if (meta && meta.next) {
+	//		getContent([lang, meta.next]);
+	//	}
+	//}, [meta && meta.next, url]);
```
rschristian, line 228: As the comment says, unfortunately lazy loading the tutorial creates a circular dep that Webpack is not a big fan of. I couldn't figure out any way to stop it from dying.

Suggestions certainly welcome if you got 'em.

src/components/app.js
```diff
@@ -1,46 +1,24 @@
-import { Component } from 'preact';
-import { Provider } from 'unistore/preact';
-import createStore from '../store';
-import Routes from './routes';
+import { LocationProvider, ErrorBoundary } from 'preact-iso';
+import { LanguageProvider } from '../lib/i18n';
+// TODO: SolutionProvider should really just wrap the tutorial,
+// but that requires a bit of refactoring
+import { SolutionProvider } from './controllers/tutorial/index.js';
```
rschristian, line 5: As noted, the issue is that we need the context _around_ the main tutorial page but at the moment kinda need the router passing the props into it. I figured this works for now, we can refactor & give it a shell later.

It might be beneficial to refactor the routing of the tutorial to sort of match the blog/blog posts.

marvinhagemeister (approved):

PR #1076, opened 2024-02-15
chore: Fix CSS "error" and remove unused complex styles

rschristian:
More stuff for Vite.

### DIFF
--- a/src/components/header/style.module.css
+++ b/src/components/header/style.module.css
@@ -481,16 +481,14 @@
 	}
 }
 
-:global {
-	.aa-dropdown-menu {
-		position: fixed !important;
-		top: 60px !important;
-		left: 50% !important;
-		transform: translateX(-50%) !important;
+:global(.aa-dropdown-menu) {
+	position: fixed !important;
+	top: 60px !important;
+	left: 50% !important;
+	transform: translateX(-50%) !important;
 
-		@media (max-width: /* --header-mobile-breakpoint */ 50rem) {
-			width: 90% !important;
-		}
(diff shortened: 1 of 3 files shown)


src/components/header/style.module.css
```diff
@@ -481,16 +481,14 @@
 	}
 }
 
-:global {
```
rschristian, line 484: Whatever Vite's using to handle CSS Modules considers this a syntax error as it drops the `:global`, resulting in a bare `{`. Easy enough to fix.

src/components/sponsors/style.module.css
```diff
@@ -40,12 +40,6 @@
 			opacity: 1;
 		}
 
-		:global(.dark) &:hover img,
-		:global(.dark) &:hover svg {
-			filter: invert(0) hue-rotate(0deg) grayscale(0)
-				drop-shadow(0 0 0.5px #fff) drop-shadow(0 5px 10px #000)
-				drop-shadow(0 0 20px rgba(255, 255, 255, 0.5));
-		}
```
rschristian, line 48: We don't use `.dark` (before my time, might've been a preference switcher that was removed at some point?), but this causes issues as it requires `:is` which is not available in the default Chrome version (87 is the default, needs 88. So close!).

What the targets should be is another topic but figured we can at least drop these unused styles that breach the defaults for now.

marvinhagemeister (approved):

src/components/sponsors/style.module.css
marvinhagemeister: Ah yeah, I think I removed the switcher at some point to make room in the header.

PR #1080, opened 2024-02-16
refactor: Reduce number of fallbacks in lambda calls

rschristian:
The release version & url had 4 different fallback values which has lead to a couple weird situations. This reduces that down.


src/components/header/gh-version.jsx
```diff
@@ -1,24 +1,22 @@
...
+		url: globalThis.prerenderPreactReleaseUrl || fallbackData.preactReleaseUrl,
+		version: globalThis.prerenderPreactVersion || fallbackData.preactVersion
+	});
 
 	useEffect(() => {
 		fetchRelease(config.repo).then(d => {
-			setVersion(d.version);
-			setUrl(d.url);
+			setRelease(d);
 		});
```
rschristian, line 14: This is a bit odd to me and I'm not quite sure if we have a bug or if it's just late and I'm being daft.

Assume we prerender for v10.19.4, generating the following HTML-ish:

```html
<a href="https://github.com/preactjs/preact/releases/tag/10.19.4">v10.19.4</a>
```

Upon hydration, say the fetch fails but catches and returns the info for `10.10.0`.

The text child of the link will be updated to `v10.10.0` as you'd expect, but the `href` will remain the same (`.../releases/tag/10.19.4`). However, if you set the URL to something wildly different before updating the state value (`d.url = 'https://google.com'; setRelease(d);`), this will update the `href`.

Are we doing something clever to try to avoid updating the attr unnecessarily?

src/prerender.js
```diff
@@ -31,9 +31,10 @@ module.exports = async () => {
 		.filter(route => !/\/:/.test(route.path))
 		.sort((a, b) => a.path.localeCompare(b.path));
 
-	let preactVersion;
 	try {
-		preactVersion = (await fetchRelease('preactjs/preact')).version;
+		const preactRelease = await fetchRelease('preactjs/preact');
+		globalThis.prerenderPreactVersion = preactRelease.version;
+		globalThis.prerenderPreactReleaseUrl = preactRelease.url;
```
rschristian, line 37: Just avoids the drilling

src/lambda/repo.js
```diff
@@ -1,5 +1,3 @@
-import fetch from 'node-fetch';
-
```
rschristian, line 2: I think this should be fine? We're running Node 20 on Netlify.

JoviDeCroock (approved):

src/lib/github.js
```diff
@@ -1,4 +1,19 @@
...
+	if (!r.ok) {
+		throw new Error(`${r.status}: Request failed for '${r.url}'`);
+	}
+	return r;
+}
+
+export const fallbackData = {
+	preactVersion: '10.19.4',
+	preactReleaseUrl: 'https://github.com/preactjs/preact/releases/tag/10.19.4',
+	preactStargazers: 35781
```
rschristian, line 15:
```suggestion
	preactVersion: '10.19.5',
	preactReleaseUrl: 'https://github.com/preactjs/preact/releases/tag/10.19.5',
	preactStargazers: 35783
```

PR #1086, opened 2024-02-19
fix: Home examples

rschristian:
`marked` is, unfortunately, turning this:

```html
<div>
    <pre repl="false"><code class="lang-jsx">
import TodoList from './todo-list';

render(&lt;TodoList /&gt;, document.body);
    </code></pre>
</div>
```

into this:

```html
<div>
    <pre repl="false"><code class="lang-jsx">
import TodoList from './todo-list';

<p>render(&lt;TodoList /&gt;, document.body);
    </code></pre></p>
</div>
```

Which breaks the code blocks pretty badly, cutting off everything after the first blank line. This is due to the fact that blank lines are an exit condition in Markdown, so the second line on out is then a paragraph.

To fix this, I've swapped out the blank lines with a `<br>`, and then in our `CodeBlock` component, made sure it could handle multiple children (as it normally just grabbed the first).

This should only impact the examples we have on the home page.

---

Additionally, I corrected the styles for the rendered examples as they were very hard to read if the user had a dark scheme preference.

### DIFF
--- a/content/de/index.md
+++ b/content/de/index.md
@@ -175,8 +175,7 @@ render(&lt;TodoList /&gt;, document.getElementById("app"));
     <div>
         <h3>Laufendes Beispiel</h3>
         <pre repl="false"><code class="lang-jsx">
-import ToDoList from './todo-list';
-
+import ToDoList from './todo-list';<br>
 render(&lt;ToDoList /&gt;, document.body);
         </code></pre>
         <div class="home-demo">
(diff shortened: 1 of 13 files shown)


src/lib/content.js
```diff
@@ -112,6 +112,8 @@ const markedWorker =
 	);
 function parseMarkdownContent(data) {
 	return markedWorker.convert(data.content).then(html => {
+		// Potentially a bug in `marked`, see: https://github.com/markedjs/marked/issues/3199
+		html = html.replaceAll('<p>render(', 'render(');
```
rschristian, line 116: Intentionally specific, to try to limit false positives.

marvinhagemeister (approved):

PR #1091, opened 2024-03-11
HOTFIX: Fix en getting-started.md

ainus64:
Se ha corregido un error en el código Markdown ya que se mostraba el símbolo # en la vista previa de la documentación.

### DIFF
--- a/content/es/guide/v10/getting-started.md
+++ b/content/es/guide/v10/getting-started.md
@@ -7,7 +7,6 @@ permalink: '/guide/getting-started'
 
 Esta guía lo ayuda a comenzar a desarrollar aplicaciones Preact. Hay 3 formas populares de hacerlo.
 
-##
 
 Si recién está comenzando, le recomendamos ir con [preact-cli](#best-practices-powered-with-preact-cli).
 


content/es/guide/v10/getting-started.md
```diff
@@ -7,7 +7,6 @@ permalink: '/guide/getting-started'
 
 Esta guía lo ayuda a comenzar a desarrollar aplicaciones Preact. Hay 3 formas populares de hacerlo.
 
-##
```
rschristian, line 10: I'm not sure this is the correct fix, this doc should (largely) match the [English version](https://github.com/preactjs/preact-www/blob/master/content/en/guide/v10/getting-started.md)

rschristian:
English commit messages would also be preferable if at all possible -- few here can read Spanish.

content/es/guide/v10/getting-started.md
ainus64: If you see spanish translated version preview, can see ## without any text heading, at first all, it's the only issue I see at the momment to read documentation.

PR #1092, opened 2024-03-15
fix: 'Run in REPL' button

### LINKED ISSUE #1079: Neither REPL example on the home page can be opened in the REPL
Reminder for myself

rschristian:
Closes #1079, upon clicking any 'Run in REPL' button, the REPL would open, but the code snippet would not be carried over. User would instead see a default or what they had open previously.

Missed in the iso migration, query params aren't passed as bare props (as they are in `preact-router`).


marvinhagemeister (approved):
Good catch!

Issue #1094, opened 2024-03-18
Broken link: component lifecycle diagram


tinymachine:
This link in `components.md` to the lifecycle diagram that Dan Abramov apparently tweeted is broken:

https://github.com/preactjs/preact-www/blob/e06b7f9eb9e0958f9b17c4a7efd8dbe14c72c32e/content/en/guide/v10/components.md?plain=1#L97

rschristian:
Looks like you can grab it from the [internet archive](https://web.archive.org/web/20191118010106/https://twitter.com/dan_abramov/status/981712092611989509), if you (or anyone else) wanted to contribute a PR fixing this.

rschristian:
Fixed in #1095

PR #1095, opened 2024-03-18
docs: update lifecycle diagram link in components.md

tinymachine:
The current link is to a deleted tweet. Update to a link to the image the author of the original tweet (Dan Abramov) posted to Imgur.

Note: I wasn't sure if the documentation should reference the old tweet, or perhaps [an archive of the tweet](https://web.archive.org/web/20191118010106/https://twitter.com/dan_abramov/status/981712092611989509) on the Internet Archive that @rschristian [found](https://github.com/preactjs/preact-www/issues/1094#issuecomment-2002939214), but I left it out for simplicity's sake. Note that Abramov doesn't seem to be referenced on the Imgur page either.


content/en/guide/v10/components.md
```diff
@@ -94,7 +94,7 @@ In order to have the clock's time update every second, we need to know when `<Cl
 | `getSnapshotBeforeUpdate(prevProps, prevState)` | called just before `render()`. return value is passed to `componentDidUpdate`.
 | `componentDidUpdate(prevProps, prevState, snapshot)` | after `render()`
 
-> See [this diagram](https://twitter.com/dan_abramov/status/981712092611989509) to get a visual overview of how they relate to each other.
+> See [this diagram](https://imgur.com/bmfcRQm) to get a visual overview of how they relate to each other.
```
rschristian, line 97: I'd say get rid of the link entirely and embed the image on the page (make sure to download the image, don't hotlink)

rschristian: You can leave mention that it's from Dan's Twitter though

tinymachine:
I embedded the image directly; not sure how PR merging on GitHub handles uploaded images, since the image now seems to be hosted in my fork. I assume the asset won't be deleted even if I eventually delete my fork if/when this PR is merged?

content/en/guide/v10/components.md
```diff
@@ -94,7 +94,9 @@ In order to have the clock's time update every second, we need to know when `<Cl
 | `getSnapshotBeforeUpdate(prevProps, prevState)` | called just before `render()`. return value is passed to `componentDidUpdate`.
 | `componentDidUpdate(prevProps, prevState, snapshot)` | after `render()`
 
-> See [this diagram](https://twitter.com/dan_abramov/status/981712092611989509) to get a visual overview of how they relate to each other.
+Here's a visual overview of how they relate to each other (originally posted in [a tweet](https://web.archive.org/web/20191118010106/https://twitter.com/dan_abramov/status/981712092611989509) by Dan Abramov):
+
+![Diagram of lifecycle methods, depicting mounting, updating, and unmounting on the x-axis, and render phase, pre-commit phase, and commit phase on the y-axis](https://github.com/tinymachine/preact-www/assets/171986/ea5fd5d1-0b9e-4125-9ca0-b94d544c0318)
```
rschristian, line 99: That's a hotlink, which we'd need to avoid. We'd want the image uploaded to this repo, in the [assets directory](https://github.com/preactjs/preact-www/tree/master/src/assets). Here's an example of how to [reference the image path](https://github.com/preactjs/preact-www/blob/e06b7f9eb9e0958f9b17c4a7efd8dbe14c72c32e/content/en/blog/introducing-signals.md?plain=1#L56) from markdown.

Exactly as you say, our docs would now depend on a) you keeping your fork around forever and b) GitHub being willing to host our image for our site (essentially paying for our bandwidth). Both are risky.

While I get and appreciate the attempt at alt text, it's better left as "Diagram of lifecycle methods" (unless you really wanted to go all-out in a description -- certainly don't feel a need to). Describing the axis, without describing the content itself, doesn't present much value. That tells them how to read the graph, but not what data is actually shown.

tinymachine: Thanks for the excellent feedback, @rschristian. I'll try to make those fixes tomorrow.

rschristian: No rush!

Thank you for not only opening an issue but also taking the time to correct it, we do really appreciate it.

rschristian (approved):
Looks great, thank you so much!

Better than it was for sure -- even if that linked still worked, it's much better having that image on the page.

tinymachine:
Thanks so much for your patience and walking me through the process, @rschristian.

PR #1104, opened 2024-03-29
fix: Update root `lang` property upon changing language

rschristian:
Supersedes #783

### DIFF
--- a/src/components/header/index.jsx
+++ b/src/components/header/index.jsx
@@ -236,8 +236,8 @@ const LanguageSelectorMenu = ({ isOpen, toggle, close, ...props }) => {
 					<use href="/assets/header-icons.svg#i18n" />
 				</svg>
 			</button>
-			<nav aria-label="submenu" aria-hidden={'' + !isOpen}>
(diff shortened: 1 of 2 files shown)


src/components/header/index.jsx
```diff
@@ -236,8 +236,8 @@ const LanguageSelectorMenu = ({ isOpen, toggle, close, ...props }) => {
 					<use href="/assets/header-icons.svg#i18n" />
 				</svg>
 			</button>
-			<nav aria-label="submenu" aria-hidden={'' + !isOpen}>
-				{Object.keys(config.languages).map(id => (
+			<nav aria-label="submenu" aria-hidden={!isOpen}>
+				{typeof window !== 'undefined' && Object.keys(config.languages).map(id => (
```
rschristian, line 240: Skip prerendering this as the "currently selected" style is going to be an issue with hydration.

We don't really need to SSR it anyhow, as it's a drop down and therefore hidden by default.

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #1100, opened 2024-03-28
fix: Ensure `useContent` hook strips fragments from URL

rschristian:
Supersedes #1099 

Easily seen by clicking the link `useContext` link beneath the [`createContext` example](https://preactjs.com/guide/v10/context#createcontext).

Scrolling to the link on navigation seems quite touchy still... trying to figure out a better system.

PR #1105, opened 2024-03-29
fix: Ensure language query params work again

rschristian:
Mostly missed in #1073, though also last night in #1100

Should use `path` in most of these cases, I blanked and didn't realize `useLocation` offers that which solves a few of our issues.

### DIFF
--- a/src/components/edit-button/index.jsx
+++ b/src/components/edit-button/index.jsx
@@ -3,10 +3,9 @@ import { useLanguage } from '../../lib/i18n';
 import style from './style.module.css';
 
 export default function EditThisPage({ isFallback }) {
-	const { url } = useLocation();
+	let { path } = useLocation();
 	const [lang] = useLanguage();
 
-	let path = url.replace(/\/$/, '') || '/index';
 	path = !isFallback ? path + '.md' : '';
 	const editUrl = `https://github.com/preactjs/preact-www/tree/master/content/${lang}${path}`;
 	return (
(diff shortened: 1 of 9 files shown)


src/lib/i18n.jsx
```diff
@@ -41,11 +41,12 @@ export function getDefaultLanguage(available, override) {
...
 
 export function LanguageProvider({ children }) {
-	const location = useLocation();
+	const { query } = useLocation();
 
 	const [lang, setLang] = useStoredValue(
 		'lang',
-		getDefaultLanguage(config.languages, location.query.lang) || 'en'
+		getDefaultLanguage(config.languages, query.lang) || 'en',
+		!!query.lang
```
rschristian, line 49: Problem with the existing implementation (IMO) is that the query param would not override a set value in localstorage, which feels wrong.

src/components/edit-button/index.jsx
```diff
@@ -3,10 +3,9 @@ import { useLanguage } from '../../lib/i18n';
 import style from './style.module.css';
 
 export default function EditThisPage({ isFallback }) {
-	const { url } = useLocation();
+	let { path } = useLocation();
 	const [lang] = useLanguage();
 
-	let path = url.replace(/\/$/, '') || '/index';
```
rschristian, line 9: The `/index` bit isn't needed as there's only two index pages on the site (`/` & `/tutorial`), neither of which show the edit button

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #1106, opened 2024-04-02
chore: Bump docsearch to v3

rschristian:
Forgive my ignorance as I'm quite unfamiliar with docsearch, but do we need to manually trigger a new crawl or something? I've noticed the content is somewhat out-of-date; Signals, for example, are entirely missing from the search

Decreases overall JS size by 24kb & CSS by 4kb.

Docsearch is no longer lazily loaded (via `requestIdleCallback`), but I'm somewhat dubious as towards the value, as it's going to be loaded quickly enough in the vast majority of situations and so another network request seems slightly wasteful. This does increase the main JS bundle by 23kb though.


src/components/header/style.module.css
```diff
@@ -131,7 +131,6 @@
 
 			&:global(.home) {
 				text-transform: uppercase;
-				padding: 0 15px;
```
rschristian, line 134: Unrelated, but I found on a small enough screen (823px as an example) our header links wrap around:

temp

Notice 'REPL' blending into our logo. Removing the extra padding helps with this, and is unnoticeable.

marvinhagemeister (approved):

PR #1108, opened 2024-04-05
refactor: Drop `netlify-lambda`, use Vite middleware for Netlify functions

rschristian:
Middleware impl mostly from @marvinhagemeister in https://github.com/preactjs/preact-www/commit/d2948f046085eff081497208b37093a110d30b0f

This also swaps out `superstatic` for `vite preview` for local testing of the prod build. Drops a unnecessary dependency, but no functional change.

I will have to alter the functions directory on Netlify, as we're no longer building them into `dist/` (as we don't need to -- Netlify should natively support everything we're using now), so they won't work in the preview. I'll need to set that right before merging.


package.json
```diff
@@ -7,13 +7,9 @@
     "node": ">=18"
   },
   "scripts": {
-    "dev": "concurrently \"npm run dev:client\" \"npm run dev:lambda\"",
-    "dev:client": "vite --port 8080",
-    "dev:lambda": "netlify-lambda serve src/lambda",
-    "start": "npm run -s server",
-    "prestart": "npm run -s build",
```
rschristian, line 14: I don't think these two were seeing any use, so I removed them.

src/lambda/repo.js
```diff
@@ -6,8 +6,8 @@ export const handler = async event => {
 		statusCode: 200,
 		body: JSON.stringify(result),
 		headers: {
-			'Cache-Control': 'public, s-maxage=1800',
-			'Access-Control-Allow-Origin': 'http://localhost:8080'
+			'Content-Type': 'application/json',
+			'Cache-Control': 'public, max-age=3600, stale-if-error=86400'
```
rschristian, line 10: Not quite sure `s-maxage` is acting as it should be anymore, it's not being cached at all. Switched to match the release function, which seems pretty reasonable.

marvinhagemeister (approved):

PR #1111, opened 2024-04-06
chore: Misc minor adjustments

rschristian:
Sorry, bit of a hodge-podge of stuff. Nothing focused/specific.


vite.config.js
```diff
@@ -46,6 +47,9 @@ function compileMarkdown(content, path) {
 
 export default defineConfig({
 	publicDir: 'src/assets',
+	optimizeDeps: {
+		include: ['@babel/polyfill', '@rollup/browser', 'sucrase']
+	},
```
rschristian, line 52: Upon opening `/repl` for the first time in dev, Vite will do a full-page refresh as detects these deps to optimize. Adding them ahead of time avoids that

src/components/controllers/repl-page.jsx
```diff
@@ -0,0 +1,14 @@
+import { Repl } from './repl';
+import { useContent } from '../../lib/use-resource';
+import { useTitle, useDescription } from './utils';
+import { useLanguage } from '../../lib/i18n';
+
+export default function ReplPage() {
```
rschristian, line 6: It was the only page left out. This fixes the i18n title as it was hard-coded before, and can allow for a bit of refactoring in the future perhaps.

src/components/controllers/doc-page.jsx
```diff
@@ -53,9 +51,8 @@ export function DocLayout() {
 
 function OldDocsWarning() {
 	const { name, version } = useRoute().params;
-	const { path } = useLocation();
 
-	if (!isDocPage(path) || version === LATEST_MAJOR) {
+	if (version === LATEST_MAJOR) {
```
rschristian, line 55: This was leading to a flash of the banner when navigating from (say) `/` -> `/guide/v10/getting-started`, as while the `path` instantly changes, `useRoute()` had a delay in delivering the route props as the router had to switch from `Page` -> `DocLayout` to `DocPage` -> `DocLayout`

src/components/controllers/doc-page.jsx
```diff
@@ -20,29 +20,27 @@ export function DocPage() {
...
-export function DocLayout() {
+export function DocLayout({ isGuide = false }) {
 	const { path } = useLocation();
 	const [lang] = useLanguage();
 
 	const { html, meta } = useContent([lang, path === '/' ? 'index' : path]);
 	useTitle(meta.title);
 	useDescription(meta.description);
 
-	const hasSidebar = meta.toc !== false && isDocPage(path);
```
rschristian, line 34: Bit redundant; `toc: false` is only set for the home page (despite `/blog`, `/about/*`, `/branding`, etc. also all making use of this component) and `isDocPage` was a regex for `^/guide`.

The `isGuide` prop is a simpler replacement, we can use the router to instead pass that information in rather than using a regex pattern.

package.json
```diff
@@ -81,9 +81,9 @@
     "marked": "^0.8.0",
     "preact": "10.15.1",
     "preact-custom-element": "^4.3.0",
-    "preact-iso": "2.2.0",
+    "preact-iso": "^2.6.2",
```
rschristian, line 84: `preact-iso` was locked as `onLoadEnd` (which we use for that progress-bar component) was never triggered w/out navigation. Refresh the page and you'd see the progress bar run forever

Fixed in `2.6.2` this morning though.

src/components/routes.jsx
```diff
@@ -5,21 +5,22 @@ import { DocPage } from './controllers/doc-page';
...
 
+// @ts-ignore
+const routeChange = url => typeof ga === 'function' && ga('send', 'pageview', url);
+
 export default function Routes() {
 	const [loading, setLoading] = useState(false);
 	return (
 		<main>
-			<progress-bar showing={loading} />
+			<loading-bar showing={loading} />
```
rschristian, line 19: IMO, "loading-bar" is a better name as we're not displaying progress per se

vite.config.js
```diff
@@ -65,8 +69,7 @@ export default defineConfig({
 					'/guide/v8/getting-started',
 					'/branding'
 				]
-			},
-			devToolsEnabled: false
+			}
```
rschristian, line 72: Unintended leftover from when Vite was a WIP, devtools work just fine

marvinhagemeister (approved):

PR #1115, opened 2024-04-09
Fix `useContext` example jsx

j-dunham:
## What is changing?
This fixes the jsx examples for `useContext` so that it is valid.  The current examples use `Foo` which is an unknown component.


marvinhagemeister (approved):
Good catch, thanks for fixing this 👍

PR #1117, opened 2024-04-09
fix: Ensure Netlify routes to `/404` when it cannot find static HTML for a page

rschristian:
Fixes two hydration issues on the 404/not found page:
  - The Preact logo/home nav in the header is missing 
  - The content overlaps the 'Edit this Page' link, making it unclickable

### DIFF
--- a/src/components/controllers/style.module.css
+++ b/src/components/controllers/style.module.css
@@ -7,7 +7,7 @@
 		margin-bottom: 4rem;
 	}
 
-	content-region:not([name='/']) :global(.markup) {
+	content-region[can-edit='true'] :global(.markup) {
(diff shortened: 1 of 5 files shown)


src/components/controllers/style.module.css
```diff
@@ -7,7 +7,7 @@
 		margin-bottom: 4rem;
 	}
 
-	content-region:not([name='/']) :global(.markup) {
+	content-region[can-edit] :global(.markup) {
```
rschristian, line 10: Not needed, and doesn't fix this issue, but more reliable

marvinhagemeister (approved):

PR #1118, opened 2024-04-09
refactor: Precompile markdown on the server

rschristian:
With our site being statically generated, we really don't have any need to compile the markdown to HTML on the client -- everything can be done ahead of time on the server, and so that's what this PR does.

It serializes to JSON, which I'm not 100% sure of. It's a marginal size increase over the raw `.md` files, but does come with the table of contents data too, so apples-to-apples, it's a tiny decrease. Might be a better format though? I did take a look at serving it as HTML directly (HTML partial sorta thing), but with the meta data JSON serialized in a script tag, it ended up being marginally larger than this (w/ gzip, anyhow)

Drops (transferred) client-side JS by ~15kb or so. Tiny decrease in render times on nav too of course.


vite.config.js
```diff
@@ -2,48 +2,13 @@ import { defineConfig } from 'vite';
...
-		content = '---\n' + JSON.stringify(meta) + '\n---\n' + content;
-	}
-
-	return content;
-}
+import { precompileMarkdown } from './plugins/precompile-markdown/index.js';
+import { netlifyPlugin } from './plugins/netlify.js';
+import { spaFallbackMiddlewarePlugin } from './plugins/spa-fallback-middleware.js';
+import { htmlRoutingMiddlewarePlugin } from './plugins/html-routing-middleware.js';
+import { rssFeedPlugin } from './plugins/rss-feed.js';
```
rschristian, line 11: Sorry for the busy diff, the config file was getting a bit big as I kept shoving more plugins/middlewares in. Needed to be split out

marvinhagemeister (approved):
This is exciting! Glad to see this being moved to server side 👍

plugins/precompile-markdown/index.js
```diff
@@ -0,0 +1,111 @@
...
+		meta
+	};
+}
+
+function markdownToHTML(data) {
+	data.html = marked(data.content);
+	return data;
+}
+
+export function applyEmojiToContent(data) {
```
rschristian, line 66:
```suggestion
function applyEmojiToContent(data) {
```

plugins/precompile-markdown/index.js
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * Parse Markdown with YAML FrontMatter into a data structure that can be reasoned about:
+ * {
+ *   content: "<html here>",
+ *   meta: { toc: [], title: "" }
+ * }
+ */
+export function parseContent(content, path) {
```
rschristian, line 28:
```suggestion
function parseContent(content, path) {
```

PR #1119, opened 2024-04-10
fix: REPL shell & code query param

rschristian:
Did this from my phone, should do the trick. I forgot we pass route props directly to the component w/ iso


marvinhagemeister (approved):
Sweet 👍

PR #1124, opened 2024-04-27
refactor: Misc accessibility improvements

rschristian:
Been trying to learn more accessibility stuff, figured I'd start here.


src/components/logo.jsx
```diff
@@ -84,7 +85,7 @@ export default class Logo extends Component {
 
 		return (
 			<svg
-				class="logo"
+				aria-label={title}
```
rschristian, line 88: `.logo` isn't used.

As for `aria-label`, it replaces a spread `title` prop (`<svg title="..."`) which does nothing to my knowledge, and is used instead of `<title>` for two reasons:

1. On hover, most browsers will show the content of `<title>` as a tooltip. We don't really need or want this in the places we use this component.
2. FireFox's accessibility tools show that `<title>` does not count as heading text in the scenario of `<h1><svg><title>...`, while `aria-label` does. This is important as we use this structure on our home page, where the main heading has no text in it, just our logo. With `<title>`, that heading is completely empty.

src/components/header/index.jsx
```diff
@@ -194,20 +194,21 @@ const NavLink = ({ to, isOpen, route, ...props }) => {
...
-		if (to.href == '/' || to.path == '/') {
-			e.preventDefault();
-			location.route('/branding');
-		}
+		e.preventDefault();
+		location.route('/branding');
 	}
+	const homeProps = to.href == '/' || to.path == '/'
+		? { onContextMenu: BrandingRedirect, 'aria-label': 'Home' }
+		: {};
```
rschristian, line 202: Better to only attach the handler to the one route it's used, and the `'aria-label': 'Home'` I think is more useful of a description, even if users are probably familiar with the concept that brand/sitename == "home"

src/style/markdown.css
```diff
@@ -395,14 +395,16 @@
...
 			a {
 				opacity: 0;
-				display: inline-flex;
 				width: 1.5rem;
 				margin-left: -1.5rem;
 
-				&::before {
-					content: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="%23fff" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M15 7h3a5 5 0 0 1 5 5 5 5 0 0 1-5 5h-3m-6 0H6a5 5 0 0 1-5-5 5 5 0 0 1 5-5h3"></path><line x1="8" y1="12" x2="16" y2="12"></line></svg>');
+				svg {
+					display: block;
```
rschristian, line 407: Not really accessibility related, but I noticed the `::before` doesn't vertically align correctly and I couldn't for the life of me fix it. No combination seemed to work.

Refactored into a `<use>` instead.

src/components/table-of-contents/index.jsx
```diff
@@ -21,7 +21,7 @@ export default function Toc() {
 
 	return (
 		<div ref={ref}>
-			<nav tabIndex={0} onFocus={this.open}>
+			<nav onFocus={this.open}>
```
rschristian, line 24: There's no real reason a user would need to tab onto the nav container, there's nothing for them to do there.

src/components/sidebar/sidebar-nav.jsx
```diff
@@ -16,10 +16,7 @@ export default function SidebarNav({ items, onClick }) {
 	const { path } = useRoute();
 
 	return (
-		<nav
-			tabIndex={0}
-			class={cx(style.toc, !(items && items.length > 1) && style.disabled)}
-		>
+		<nav class={cx(style.toc, !(items && items.length > 1) && style.disabled)}>
```
rschristian, line 19: Same here, tabbing onto the side bar has no use.

src/components/markdown.jsx
```diff
@@ -1,27 +0,0 @@
-import marked from 'marked';
```
rschristian, line 1: Sorry, unrelated, but long unused.

src/components/content-region/index.jsx
```diff
@@ -20,11 +20,19 @@ const COMPONENTS = {
...
-	const Tag = 'h' + i;
+// Skip h1, no need to link to a page title
+for (let i = 2; i <= 6; i++) {
+	const Tag = /** @type {keyof preact.JSX.IntrinsicElements} */ ('h' + i);
 	COMPONENTS[Tag] = function header(props) {
 		props.children = props.id
-			? [<a href={'#' + props.id} />, props.children]
+			? [
+				<a href={'#' + props.id}>
+					<svg width="16" height="16" viewBox="0 0 24 24" aria-label={`Link to: ${props.children} (#${props.id})`}>
```
rschristian, line 30: This implementation (e.g., `Link to: Setting up JSX (#setting-up-jsx)`) came from Vue's docs which I quite like, makes it quite clear why there's a link right alongside the header content.

src/components/code-editor/code-mirror.css
```diff
@@ -58,7 +58,7 @@
 	}
 
 	.CodeMirror-linenumber {
-		color: #4b5363;
+		color: #8b93a3;
```
rschristian, line 61: Was failing contrast checks

marvinhagemeister (approved):
Good stuff 👍

PR #1125, opened 2024-04-27
fix: broken image on Components page

tinymachine:
I previously added this image, but it seems it's broken, maybe because assets no longer are (or never were?) contained in a `/assets` directory. Removing `/assets` from the image `src` should fix this.


rschristian (approved):
Oops, thank you. I had forgotten to re-check assets paths when finishing up our Vite migration and rebasing my branch.

PR #1129, opened 2024-05-06
fix: logo component type error and compatible with Windows

### LINKED ISSUE #1128: Can't run site in development mode
The site can't run on my local machine. All things I do:

1. Clone this repo
2. Run `npm i` and `npm run dev`
3. Open localhost:8080

Then I got an error in the console as follows image and no contents rendered:

image

But it looks like the build output works fine, there are no error throws during building at least.

My Node.js version is 18.19.1

KiritaniAyaka:
Fix #1128


rschristian (approved):
Thanks!!

PR #1130, opened 2024-05-07
refactor: Swap the OG image to one including the Preact name

rschristian:
Just playing with this a bit for now


JoviDeCroock (approved):
Amazing

marvinhagemeister (approved):

rschristian:
There is admittedly a loss of "at a glance" information which I'm growing a bit conflicted on. Not all of our descriptions are great but I'm not sure if removing it outright is really adding anything (even if the full branding looks marginally better, IMO).

Might let this sit for a bit.

PR #1131, opened 2024-05-08
refactor: Hide banner on short viewports (like rotated phones)

### LINKED ISSUE #847: Tutorial is hardly usable on mobile due to the huge header
I have to use my mobile phone for learning Preact, but the way it's being rendered leaves the content barely a half of the screen:

Screenshot_20220509-164353_Chrome

As you see, there are two culprits. 

First is the page header, which allocates the bigger part of the upper half.

The second one is the browser address bar which never gets hidden due to the way the left column is scrolled. I believe the page header cannot be hidden on the same reason - no body scroll.

So if that scroll is the real issue here, I'd support removing the scroll from the left container, at least on mobile.

rschristian:
Closes #847

First two are from this PR preview, last is from the linked issue:

With hidden toolbar |  Without | Previous State
:-------------------------:|:-------------------------:|:-------------------------:
image | image | Image

Not sure if all mobile browsers can hide their toolsbars (that's Safari from me), but we've at least reclaimed a good portion of screen real estate. Maybe we could do some sort of expanding/collapsing header too (or in place of this), but that'd need some more thought & work.

Hopefully this isn't problematic, it's only hiding the banner in a pretty specific situation when vertical height is pretty badly needed.

### DIFF
--- a/src/components/controllers/tutorial/style.module.css
+++ b/src/components/controllers/tutorial/style.module.css
@@ -18,11 +18,17 @@
 .tutorial {
 	position: absolute;
 	left: 0;
-	top: 96px;
+	top: var(--header-and-banner-height);
 	bottom: 0;
 	height: auto;
 	width: 100%;
 	overflow: hidden;
+
+	@media (max-height: 431px) {
+		& {
+			top: var(--header-height);
+		}
+	}
 
 	:global(.markup) {
 		p:first-of-type {
--- a/src/components/header/style.module.css
+++ b/src/components/header/style.module.css
@@ -11,6 +11,12 @@
 	b {
 		color: #ffd600;
 	}
+
+	@media (max-height: 431px) {
+		& {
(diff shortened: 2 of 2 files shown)


src/components/controllers/tutorial/style.module.css
```diff
@@ -18,12 +18,18 @@
...
 	position: absolute;
 	left: 0;
-	top: 96px;
+	top: var(--header-and-banner-height);
 	bottom: 0;
 	height: auto;
 	width: 100%;
 	overflow: hidden;
 
+	@media (max-height: 431px) {
```
rschristian, line 27: 432px seemed like a reasonably safe value, just above the width of most phones (going by the device emulator in my devtools). The vast majority of phones should fall below this, giving them the little bit of extra height when rotated.

marvinhagemeister (approved):

PR #1133, opened 2024-05-19
Fix `useId` example comment

foxt:
Currently the comment reads `// Display a nice error message`, which doesn't make sense, seems to be copied from the `useErrorBoundary` example?


rschristian (approved):
Good catch, thanks!

PR #1136, opened 2024-05-26
refactor: Drop 'Progressive Web Apps' guide page

rschristian:
It's not really a useful page, mostly pushing users to Preact-CLI and not giving enough useful or actionable info for users to make a PWA with Preact.

This information, what little of it is there, is better off hosted and maintained elsewhere.


marvinhagemeister (approved):
Good point, I've been thinking this for a while too.

PR #1139, opened 2024-05-27
Blog post about the 5 years of Preact X

JoviDeCroock:


rschristian (approved):
Great work and idea to write about this!

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,141 @@
...
+
+A lot of you have been waiting for [Preact 11](https://github.com/preactjs/preact/issues/2621), announced in an issue opened
+way back in July 2020, and to be clear I was one of the most excited people for v11.
+When we started thinking about Preact 11 we believed that there was no way to introduce the changes we had in mind
+in Preact X without breaking changes, some of the things we had in mind:
+
+- Using a backing VNode structure to reduce GC, by doing this we'd only use the result of `h()` to update our backing-node.
+- Reconciler performance, adding fast paths for mounting/...
+- Some changes like removing `px` suffixing, `forwardRef` and breaking IE11 support.
+- Addressing event/child diffing bugs.
```
marvinhagemeister, line 18: Should we add the `keeping ref in props` change here. We had to backtrack to react compat.

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,141 @@
...
+We have faced a number of issues where an event could bubble up to a newly inserted VNode which would result in undesired behaviour, this was fixed
+[by adding an event-clock](https://github.com/preactjs/preact/pull/4322). In the following scenario, you would click the button which sets state, the browser
+interleaves event bubbling with micro-ticks, which is also what Preact uses to schedule updates. This combination means that Preact will update the UI, meaning
+that the `<div>` will get that `onClick` handler which we'll bubble up to and invoke the `click` again toggling this state immediately off again.
+
+```jsx
+const App = () => {
+  const [toggled, setToggled] = useState(false);
+
+  return toggled ? <div onClick={() => setToggled(false)}><span>clear</span></div> : <div><button onClick={() => setToggled(true)}>toggle on</button></div>
```
marvinhagemeister, line 113: Nit: let's format this to be more readable

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,141 @@
...
+these experiments for Preact 11, we started noticing that the improvements we were landing didn't need to be exclusive to Preact 11.
+
+## Releases
+
+Since the aforementioned Preact 11 issue there have been 18 (!!) minor versions of Preact X.
+Many of them have been directly inspired by work done on Preact 11. Let's go over a few and look at the impact.
+
+### 10.5.0
+
+The introduction of [resumed hydration](https://github.com/preactjs/preact/pull/2754) this functionality basically allows suspending during
```
rschristian, line 36:
```suggestion
The introduction of [resumed hydration](https://github.com/preactjs/preact/pull/2754) -- this functionality basically allows suspending during
```

Semicolon would work too I think, but needs something to break up the flow of the sentence.

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,141 @@
...
+
+We value our ecosystem and we value the extensions being built through our [`options API`](https://marvinh.dev/blog/preact-options/),
+this is one of the main drivers behind not wanting to introduce these breaking changes but instead, allow you all to benefit
+from our research without a painful migration path.
+
+This doesn't mean that Preact 11 won't happen but it might not be the thing that we initially thought it would be, instead we might just drop IE11 support and give you
+those performance improvments, all while giving you the stability of Preact X. We could instead work on a meta-framework for Node to accompany Deno as a great
+way to write Preact, we could bet more on signals... There are a lot of options while avoiding leaving a part of the community behind.
+
+We aren't gone, we are here to listen to you and we want to grow this community together.
```
marvinhagemeister, line 141: I don't have a better suggestion for the phrasing yet, but this reads a bit odd to me.

1. Worried that folks think that Preact is mostly for Deno going onwards

> There are a lot of options while avoiding leaving a part of the community behind.
This feels like our main worry is not to be left behind? Do we feel that way?

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,150 @@
...
+- [A new async router](https://github.com/preactjs/preact-iso)
+- [Create Preact](https://github.com/preactjs/create-preact)
+
+We value our ecosystem and we value the extensions being built through our [`options API`](https://marvinh.dev/blog/preact-options/),
+this is one of the main drivers behind not wanting to introduce these breaking changes but instead, allow you all to benefit
+from our research without a painful migration path.
+
+This doesn't mean that Preact 11 won't happen but it might not be the thing that we initially thought it would be, instead we might just drop IE11 support and give you
+those performance improvments, all while giving you the stability of Preact X. We could instead work on a meta-framework for Node like [Fresh](https://fresh.deno.dev/) as a great
+way to write Preact, we could bet more on signals... There are a lot of options while avoiding leaving a part of the community behind.
```
marvinhagemeister, line 150: Minor rewording.

```suggestion
This doesn't mean that Preact 11 won't happen but it might not be the thing that we initially thought it would be. Instead, we might just drop IE11 support and give you
those performance improvements, all while giving you the stability of Preact X. There are many more ideas floating around and we're very interested in the wider Preact experience in the context of meta-frameworks that provide things like routing out of the box. We're exploring this angle in our vite preset as well as [Fresh](https://fresh.deno.dev/) to get a good feel what a Preact first meta framework should look like.
```

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,150 @@
...
+
+# Preact X, a story of stability
+
+A lot of you have been waiting for [Preact 11](https://github.com/preactjs/preact/issues/2621), announced in an issue opened
+way back in July 2020, and to be clear I was one of the most excited people for v11.
+When we started thinking about Preact 11 we believed that there was no way to introduce the changes we had in mind
+in Preact X without breaking changes, some of the things we had in mind:
+
+- Using a backing VNode structure to reduce GC, by doing this we'd only use the result of `h()` to update our backing-node.
+- Reconciler performance, adding fast paths for mounting/...
```
marvinhagemeister, line 16: Minor rewording:

```suggestion
- Reconciler performance, to allow optimized paths for mounting/unmounting/etc
```

content/en/blog/preact-x.md
```diff
@@ -0,0 +1,150 @@
...
+- Some changes like removing `px` suffixing, `forwardRef` and breaking IE11 support.
+- Keeping ref in props.
+- Addressing event/child diffing bugs.
+
+The above were our goals, when breaking these down we can clearly see that there's a distinction here, only point 3 would
+be a breaking change for our consumers while point 1 would have a huge impact to folks building extensions like Preact Signals,
+Preact devtools, ... This could break a lot of the cool stuff our community is working on to extend Preact.
```
marvinhagemeister, line 23:
```suggestion
These were our initial goals for v11, but upon going down this path, we realised that many of those changes weren't actually breaking changes and could be released directly in v10 in a non-breaking way. Only the third point, removing the `px` suffix and passing `ref` directly in props as well as dropping IE11, fall into the breaking changes category. We went with releasing the other features in the stable v10 release line, which allows any Preact user to benefit from them immediately, without having to change their code.

Preact has a much bigger userbase today compared to when we made the original plans for v11. It enjoys wide usage in many small to big companies for mission critical software. We really want to be sure that any breaking changes we may introduce, are absolutely worth the cost of moving the whole ecosystem over to it.
```

marvinhagemeister (approved):
Did a last pass and I think it's ready to be published after that 👍

content/en/blog/preact-x.md
rschristian: That last sentence in this suggestion should drop the comma I think, breaks the flow.

marvinhagemeister: Good point, updated.

PR #1148, opened 2024-06-18
chore: Bump codemirror

rschristian:
Bumps CodeMirror up to v6, somewhat pulled from #793 (man was that PR a gold mine of stuff). Bit of an increase on the bundle size (loading the 'Simple Counter' w/ no cache: `393kb` -> `448kb`, `-12kb` for the uncompressed favicon in preview, or ~43kb bump), but I don't think it's too bad.

I've enabled auto completion (basic, but not bad) and bracket matching with this bump, seem like general nice-to-haves and I am already quite happy with these. There seems to be quite a bit more we can also enable/play around with in the future too.


src/components/controllers/tutorial/index.jsx
```diff
@@ -134,15 +128,14 @@ export class Tutorial extends Component {
 		this.setState({ error: null });
 	};
 
-	render({ html, meta, loading }, { code, error }) {
+	render({ html, meta, loading: routerLoading }, { code, error, loading: tutorialLoading }) {
```
rschristian, line 131: Drive-by loading mechanism improvement over #1143 

Still needs work, but incorporates both routing (markdown and tutorial bundles) with the tutorial orchestration (editor & repl runner) better.

src/components/controllers/repl/index.jsx
```diff
@@ -111,26 +113,17 @@ export class Repl extends Component {
 	}
 
 	componentDidMount() {
-		Promise.all([
-			import('../../code-editor'),
-			import('./runner')
-		]).then(([CodeEditor, Runner]) => {
```
rschristian, line 117: No real need to lazily init like this, Vite is smart enough to combine the code editor & runner imports so that tutorial & repl can share them, which is nice.

Might need to keep an eye on that though. Might also revert yet, dunno. Thought I'd try it anyhow.

src/components/code-editor/index.jsx
```diff
@@ -1,99 +1,72 @@
...
-	componentWillUnmount() {
-		let wrapper = this.editor && this.editor.getWrapperElement();
-		if (wrapper) this.base.removeChild(wrapper);
-		this.editor = null;
-	}
+		return () => {
+			view.destroy();
+			setEditor(null);
+		};
+	}, [props.baseExampleSlug]);
```
rschristian, line 69: TODO: This probably needs some work.

src/components/code-editor/index.jsx
```diff
@@ -1,99 +1,72 @@
...
-		if (this.errors) {
-			this.editor.operation(() => {
-				this.errors.forEach(e => e.clear());
-			});
-			this.errors.length = 0;
-		}
-
-		if (!error || !error.loc) return;
+// Custom theme that better matches our Prism config, though
+// the lexer is somewhat limited so it still deviates
```
rschristian, line 16: Theme definitely isn't perfect, but this is a bit closer to our Prism config than the previous set up was.

src/components/controllers/repl/index.jsx
```diff
@@ -308,15 +306,6 @@ export class Repl extends Component {
...
 		<loading-bar showing={!!loading} />
-		<style>{`
-			main {
-				height: 100% !important;
-				overflow: hidden !important;
-			}
-			footer {
-				display: none !important;
-			}
-		`}</style>
```
rschristian, line 319: Drive-by change, but I see no reason why this needs to be a style tag?

src/components/code-editor/index.jsx
```diff
@@ -1,99 +1,106 @@
...
+				bracketMatching(),
+				autocompletion(),
+				javascript({ jsx: true }),
+				keymap.of([indentWithTab, ...defaultKeymap, ...historyKeymap]),
+				[theme, syntaxHighlighting(highlightStyle, { fallback: true })],
+				EditorView.updateListener.of(update => {
+					// Ignores changes from swapping out the editor code programmatically
+					if (isViewUpdateFromUserInput(update)) {
+						props.onInput(update.state.doc.toString());
+					}
```
rschristian, line 78: TODO: Upon further testing, this seems to be a race condition. Any input triggers two events, one in which `isViewUpdateFromUserInput()` returns `true`, the second, `false`. Which has the updated doc string is inconsistent.

src/components/controllers/repl-page.jsx
```diff
@@ -38,42 +38,31 @@ export default function ReplPage() {
  *
  * ?code -> ?example -> localStorage -> simple counter example
  */
-function initialCode(query) {
-	let code, slug;
```
rschristian, line 42: TODO: Split this out. Needs to land still, reduces the mess, but not necessarily needed here.

src/components/code-editor/index.jsx
```diff
@@ -1,99 +1,106 @@
...
-					<div ref={r => (ref = r)} class={style.lintError}>
-						<pre style={`padding-left:${left}px;`}>▲</pre>
-						<div>🔥 {error.message.split('\n')[0]}</div>
-					</div>
-				);
-				render(errorLine, this.scratch);
-				this.errors = [this.editor.addLineWidget(error.loc.line - 1, ref)];
-			});
-		}, 1000);
-	}
```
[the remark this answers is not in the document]
rschristian: Actually, the tutorial does seem to show errors. Will need to take another look.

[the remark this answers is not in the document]
rschristian: The error positioning is pretty touchy & limited, we can't effectively map the built output back to the code editor input.

Redoing the compilation is a can of worms I don't want to address here. For now, I've disabled the error overlay in the editor as it barely works and tends to point at the wrong line entirely.

JoviDeCroock (approved):

PR #1155, opened 2024-06-27
refactor: Better utilize suspense in Tutorial

rschristian:
This should be the last of the vite & preact-iso migration changes, sorry it took so long for me to land.

Better utilizes suspense, so now we can just rely on the router loading state -- no longer is there any loading of the REPL/Code Editor separately post-mount. This also fixes the full unmount of the route when navigating between tutorial steps, structure is now preserved & content is simply swapped out. Feels a lot smoother.

Certainly let me know if any issues are spotted -- I spent a fair bit of time going through different situations manually but this is the most complex page we have by a fair margin, which is why it was left for last.

### DIFF
--- a/src/components/controllers/tutorial/style.module.css
+++ b/src/components/controllers/tutorial/style.module.css
@@ -242,18 +242,6 @@
 			transition: opacity 1s ease;
 			overflow: auto;
 
-			.error {
-				position: absolute;
-				left: 0;
-				right: 0;
-				bottom: 0;
-				margin: 0;
-				height: auto;
-				border-top: 2px solid var(--color-error-heading);
-				padding: 10px;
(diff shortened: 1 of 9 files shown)


src/components/controllers/tutorial/style.module.css
```diff
@@ -1,20 +1,3 @@
...
-	display: flex;
-}
-
-.close {
-	position: absolute;
-	top: 1rem;
-	right: 1rem;
-	z-index: 10;
-}
-
```
rschristian, line 17: Most of these error styles weren't working correctly as they were sitting under the REPL Runner's styles and/or were redundant, so I stripped them out. Errors should continue to look as they have in the REPL & Tutorial.

JoviDeCroock (approved):

PR #1156, opened 2024-06-29
Update web-components.md to explain camelCase vs dash-case props for custom elements

trusktr:
ref:

- https://github.com/preactjs/preact/issues/4418

### DIFF
--- a/content/en/guide/v10/web-components.md
+++ b/content/en/guide/v10/web-components.md
@@ -50,6 +50,12 @@ function Foo() {
 }
 ```
 
+> [!Note]
+> Preact maps camelCase JSX property names to the JS properties of the custom elements, not dash-case
+> like other other JSX compilers or runtimes may. If a custom element has a property names `someProperty`, using
+> the syntax `someProperty={value}` will work while `some-property={value}` will not (the latter will
+> set the attribute instead).
+
 When rendering static HTML using `preact-render-to-string` ("SSR"), complex property values like the object above are not automatically serialized. They are applied once the static HTML is hydrated on the client.
 
 ### Accessing Instance Methods
--- a/content/en/guide/v10/web-components.md
+++ b/content/en/guide/v10/web-components.md
@@ -50,11 +50,7 @@ function Foo() {
 }
 ```
 
-> [!Note]
-> Preact maps only camelCase JSX property names to the same camelCase JS properties of the custom elements,
-> and does not map dash-case JSX properties to camelCase JS properties like other JSX compilers or runtimes might.
-> If a custom element has a property named `someProperty`, using the syntax `someProperty={value}` will work while
-> `some-property={value}` will not (the latter will set the attribute instead).
+> **Note:** Preact makes no assumptions over naming schemes and will not attempt to coerce names, in JSX or otherwise, to DOM properties. If a custom element has a property name `someProperty`, it will need to be set using `someProperty=...` rather than `some-property=...`.
 
 When rendering static HTML using `preact-render-to-string` ("SSR"), complex property values like the object above are not automatically serialized. They are applied once the static HTML is hydrated on the client.
 
--- a/content/en/guide/v10/web-components.md
+++ b/content/en/guide/v10/web-components.md
@@ -51,10 +51,10 @@ function Foo() {
 ```
 
 > [!Note]
-> Preact maps camelCase JSX property names to the JS properties of the custom elements, not dash-case
-> like other other JSX compilers or runtimes may. If a custom element has a property names `someProperty`, using
-> the syntax `someProperty={value}` will work while `some-property={value}` will not (the latter will
-> set the attribute instead).
+> Preact maps only camelCase JSX property names to the same camelCase JS properties of the custom elements,
+> and does not map dash-case JSX properties to camelCase JS properties like other JSX compilers or runtimes might.
+> If a custom element has a property named `someProperty`, using the syntax `someProperty={value}` will work while
+> `some-property={value}` will not (the latter will set the attribute instead).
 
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/web-components.md
```diff
@@ -50,6 +50,12 @@ function Foo() {
 }
 ```
 
+> [!Note]
+> Preact maps only camelCase JSX property names to the same camelCase JS properties of the custom elements,
+> and does not map dash-case JSX properties to camelCase JS properties like other JSX compilers or runtimes might.
+> If a custom element has a property named `someProperty`, using the syntax `someProperty={value}` will work while
+> `some-property={value}` will not (the latter will set the attribute instead).
```
trusktr, line 57: > [!Note]
> The `[!Note]` syntax results in a note like this one.

rschristian, line 57:
```suggestion
> **Note:** Preact makes no assumptions over naming schemes and will not attempt to coerce names, in JSX or otherwise, to DOM properties. If a custom element has a property name `someProperty`, it will need to be set using `someProperty=...` rather than `some-property=...`.
```

Slightly different note syntax that I think our docs, by and large anyhow, prefers.

I'd skip over the bits about other frameworks or compilers, it's not particularly useful information. We'll also skip over the "will set the attribute instead" comment as that's not guaranteed, depending on the circumstances (albeit, it'd be an oddity).

rschristian: I thought it did, but appears not to: https://deploy-preview-1156--preactjs.netlify.app/guide/v10/web-components#properties-and-attributes

trusktr: >  We'll also skip over the "will set the attribute instead" comment as that's not guaranteed, depending on the circumstances (albeit, it'd be an oddity).

That's curious. Do you mean, for example, if the element does have a `foo-bar` JS property then that won't set the attribute? Or is there another case?

rschristian: Precisely, yes.

The logic largely boils down to:

```js
if (!somethingThatNeedsToBeSetAsAnAttr && prop in domElement) domElement.prop = ...
else domElement.setAttribute(prop, ...)
```

So if `domElement['some-property']` exists, Preact will happily update the property, rather than falling through to the attribute.

rschristian (approved):
Thanks for adding this

trusktr:
No prob! Still curious what you meant here, to understand if there's something to document, or just that `foo-bar` could be set as an attribute unless there's an actual `foo-bar` JS propert (which is rare).

https://github.com/preactjs/preact-www/pull/1156#discussion_r1659903818

rschristian:
> or just that `foo-bar` could be set as an attribute unless there's an actual `foo-bar` JS propert (which is rare).

More that "`foo-bar` **will** be set as an attribute...", but yes, just this.

In your original bug on the Preact repo, if you defined your methods as `get 'mount-point'()` & `set 'mount-point'()`, all would have worked. For any old DOM node, you could use [option hooks](https://preactjs.com/guide/v10/options) to add arbitrary props before diffing (alternatively `useEffect()` & a `useRef` certainly does the trick too), but that'd indeed be pretty rare to see.

trusktr:
> More that "`foo-bar` **will** be set as an attribute...", but yes, just this.

Just to confirm I understand, `foo-bar` will set as an attribute regardless if there's a `foo-bar` JS property? If so, this would be a useful tidbit too.

trusktr:
> No `foo-bar` property means it ends up as an attribute; it's an either-or situation (not counting reflection).

Ah, so 

- no `foo` property means it will end up as a JS property anyway,
- but no `foo-bar` property means it will end up as an attribute

Is that it?

rschristian:
I'm not sure how you're getting confused here, have I made a typo somewhere? As I've written a few times now, it's as simple as:

```js
if (!somethingThatNeedsToBeSetAsAnAttr && 'foo-bar' in domElement) domElement['foo-bar'] = ...
else domElement.setAttribute('foo-bar', ...)
```

If you have `foo-bar`, and it doesn't exist as a property upon the element, it'll be set as an attribute. This is all we're doing.

> no `foo` JS property means it will end up as a JS property anyway,

`foo` is completely irrelevant to the setting of `foo-bar`. They're different names and have 0 interaction whatsoever.

trusktr:
My confusion was because you mentioned that,

> We'll also skip over the "will set the attribute instead" comment as that's not guaranteed, depending on the circumstances (albeit, it'd be an oddity).

which implied there is some behavior not guaranteed depending on some circumstances, so I was looking for certain explanation to eliminate any assumptions.

> foo is completely irrelevant to the setting of foo-bar. They're different names and have 0 interaction whatsoever.

Yeah totally, I know that.

Sorry, so just to confirm, when two JSX properties such as `foo` and `lorem-ipsum` both do not fall into the "tiny list of exceptions", both get treated with the _same exact_ conditional checks?

I just want to make sure we've gotten 100% clarity for the docs.

rschristian:
>> We'll also skip over the "will set the attribute instead" comment as that's not guaranteed, depending on the circumstances (albeit, it'd be an oddity).
>
> to me it implied there was some behavior not guaranteed depending on some circumstances, so I was looking for a certain explanation to eliminate any assumptions.

Ah, gotcha. Yeah I meant that it's not guaranteed `some-property` (from the example) will be set as an attribute; if, for some reason, `some-property` exists as a property on the DOM element, we will update/set the property. That would be fairly uncommon but not impossible.

> So just to confirm finally, when two JSX properties such as `foo` and `lorem-ipsum` both do not fall into the "tiny list of exceptions", both get treated the same.
> 
> regardless of the property name (i.e. dashes make absolutely no difference).

Yep, that's the intention.

PR #1157, opened 2024-06-30
refactor: Clean up a bit post #1155 & #1150

rschristian:
Few things I ran into in #1148 that can be extracted a bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "rules": {
       "react/sort-comp": 0,
       "react/no-danger": 0,
+      "react/jsx-no-bind": 0,
       "brace-style": 0,
       "indent": 0,
       "lines-around-comment": 0,
--- a/src/components/controllers/repl-page.jsx
+++ b/src/components/controllers/repl-page.jsx
@@ -1,6 +1,6 @@
-import { useRoute } from 'preact-iso';
+import { useLocation, useRoute } from 'preact-iso';
 import { Repl } from './repl';
-import { useExample } from './repl/examples';
+import { fetchExample } from './repl/examples';
 import { useContent, useResource } from '../../lib/use-resource';
 import { useTitle, useDescription } from './utils';
 import { useLanguage } from '../../lib/i18n';
@@ -15,7 +15,7 @@ export default function ReplPage() {
 	useTitle(meta.title);
 	useDescription(meta.description);
 
-	const [code, slug] = initialCode(query);
+	const code = useResource(() => getInitialCode(query), [query]);
 
 	return (
 		<div class={style.repl}>
@@ -24,11 +24,8 @@ export default function ReplPage() {
 					height: 100% !important;
 					overflow: hidden !important;
 				}
-				footer {
-					display: none !important;
-				}
 			`}</style>
-			<Repl code={code} slug={slug} />
+			<Repl code={code} />
 		</div>
 	);
 }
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "rules": {
       "react/sort-comp": 0,
       "react/no-danger": 0,
+      "react/jsx-no-bind": 0,
```
rschristian, line 23: Is anyone against this getting silence? Granted, many of them are my fault, but we have dozens of warnings that I don't think are important to anyone these days?

src/components/controllers/tutorial/index.jsx
```diff
@@ -140,42 +143,38 @@ export function Tutorial({ html, meta }) {
 					orientation="horizontal"
 					force={!showCode ? '100%' : undefined}
 					other={
-						// TODO: CodeMirror v5 cannot load in Node, and loading only the runner
-						// causes some bad jumping/pop-in. For the moment, this is the best option
-						typeof window === 'undefined'
-							? null
-							: <Splitter
```
rschristian, line 147: There is already a `if (typeof window === 'undefined')` return above, this does nothing. Was a relic from when I was trying to see if I could prerender the content portion of the tutorial, but skip the editor & runner (I could not, at least not well).

Evidently I forgot to clean it up.

src/components/controllers/repl-page.jsx
```diff
@@ -38,42 +35,31 @@ export default function ReplPage() {
  *
  * ?code -> ?example -> localStorage -> simple counter example
  */
-function initialCode(query) {
-	let code, slug;
+async function getInitialCode(query) {
+	const { route } = useLocation();
```
rschristian, line 39: Admittedly this block was a bit of a mess that I should have given a better look over.

Fixes an issue with `?code` being cleared on mount (though it would populate correctly) and drops `history.replaceState` in a few places in favor of `route()`.

This still isn't great, `preact-iso` has a rather fundamental issue at the moment (IMO) in that it updates the location context as soon as it matches a route. This leads to really awful control flow w/ suspense as you cannot rely on the content being updated alongside a path change -- content will update in a subsequent render. It's workable, just... fiddly. I was trying to avoid some of the woes by using `history.replaceState` but this refactor should work well while getting us in a position to take advantage of a fix to `iso` in the future (though I think it'll need a major).

src/components/controllers/repl/index.jsx
```diff
@@ -13,55 +14,49 @@ import REPL_CSS from './examples.css?raw';
  * @param {string} props.code
  * @param {string} [props.slug]
  */
-export function Repl({ code, slug }) {
+export function Repl({ code }) {
+	const { route } = useLocation();
+	const { query } = useRoute();
 	const [editorCode, setEditorCode] = useStoredValue('preact-www-repl-code', code);
-	const [exampleSlug, setExampleSlug] = useState(slug || '');
```
rschristian, line 18: Most of the slug changes here are, again, due to me trying to run around `iso` when I wrote this due to the awkward update order.

src/components/controllers/repl-page.jsx
```diff
@@ -24,11 +24,8 @@ export default function ReplPage() {
 					height: 100% !important;
 					overflow: hidden !important;
 				}
-				footer {
-					display: none !important;
-				}
```
rschristian, line 29: No footer on the REPL of course -- I had just copy/pasted this CSS, it's not relevant anymore.

marvinhagemeister (approved):

PR #1160, opened 2024-07-01
refactor: Only debounce REPL runner on user input

### LINKED ISSUE #1159: Limit runner delay to user input
The REPL runner has a 500ms timeout on code change, which is well and fine for user input (a tad high for my liking perhaps, but avoids noisy & useless errors) but less so when say switching REPL examples or tutorial pages. On the REPL especially, as some examples need to pull additional modules from esm.sh, this feels rather laggy and slow.

https://github.com/preactjs/preact-www/blob/fa8a917e9804292344f1da68794833f1a0c15b36/src/components/controllers/repl/runner.jsx#L69-L85

Off the top of my head, probably best to move this timeout into consumers of the REPL runner, let them control the timeouts as they can better discern user input from controlled swapping of runner code.

rschristian:
Closes #1159

Instead of debouncing the REPL runner by 500ms across the board, limit it to user interaction. This lets us switch examples in the REPL or pages in the tutorial faster -- the former is especially relevant, as dynamically pulling modules from `esm.sh` means there often will be a delay anyhow. Decreasing the baseline delay is therefore quite useful to make the editor seem faster & more fluid.

### DIFF
--- a/src/components/controllers/repl/index.jsx
+++ b/src/components/controllers/repl/index.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'preact/hooks';
+import { useEffect, useState } from 'preact/hooks';
 import { useLocation, useRoute } from 'preact-iso';
 import { Splitter } from '../../splitter';
 import { EXAMPLES, fetchExample } from './examples';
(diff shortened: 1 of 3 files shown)


src/components/controllers/repl/index.jsx
```diff
@@ -52,6 +54,13 @@ export function Repl({ code }) {
 		}
 	};
 
+	useEffect(() => {
+		const delay = setTimeout(() => {
+			setRunnerCode(editorCode);
+		}, 250);
```
rschristian, line 60: Happy to revert, but 250ms felt a bit better? This only affects the REPL runner pane, not the editor itself, so it's not displaying errors that get in the user's way.

Just a bit more responsive for simple edits is all.

marvinhagemeister (approved):

PR #1164, opened 2024-07-03
refactor: Better align REPL examples for comparison

rschristian:
We have a couple examples that beg comparison (`counter` vs `counter-htm`, `todo-list` vs `todo-list-signals`) but have implemented them rather differently so that the comparison mostly falls flat -- it's very much an apples-to-oranges comparison.

While that's not inherently a problem, I think it's better to more closely align these examples so that we can better show newer users the direct translation of some (very basic) concepts of these APIs we offer.

I might go pick up a list of other common examples that other framework show, might be nice to flesh this list out further.

I also ran all the examples through Prettier, some were a bit inconsistent


marvinhagemeister (approved):

src/components/controllers/repl/examples/todo-lists/todo-list-signals.txt
```diff
@@ -0,0 +1,58 @@
...
+							{todo.completed ? <s>{todo.text}</s> : todo.text}
+						</label>{' '}
+						<button onClick={() => removeTodo(index)}>❌</button>
+					</li>
+				))}
+			</ul>
+			<p>Completed count: {completedCount.value}</p>
+		</form>
+	);
+}
```
ushuz, line 56: The updated example doesn't work. Try removing a todo.

rschristian: Seems to work just fine for me? What issue are you seeing?

ushuz: The remove button doesn't actually remove the todo item, it only empties the text. The previous example works just fine.

preactjs

rschristian: Strange, I can't reproduce. Try refreshing whilst clearing the cache, you might have something left over? Else, I'll look into it when I can.

https://github.com/user-attachments/assets/0117094d-9c3e-4351-beb7-e4d1077b5b68

ushuz: Refreshing doesn't help in my case. My browser is Chrome 128.0.6613.114 on Windows 10.

I made a longer recording, and based on my observation, it seems that clicking the remove button indeed removed the todo item, but it also triggered the form submission and added an empty todo item:

preactjs

rschristian: Ah, yep, forgot `type="button"` as per usual.

Good catch!

PR #1166, opened 2024-07-16
docs: Update controlled component example

rschristian:
re: post in slack, people might find this example confusing so we should probably just cut it. I moved some sentences around a bit, I think it's more clear? Not sure, let me know!

### DIFF
--- a/content/en/guide/v10/forms.md
+++ b/content/en/guide/v10/forms.md
@@ -17,11 +17,18 @@ The main difference is that in most cases the `value` is not controlled by the D
 
 ## Controlled & Uncontrolled Components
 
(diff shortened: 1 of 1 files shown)


content/en/guide/v10/forms.md
```diff
@@ -17,21 +17,18 @@ The main difference is that in most cases the `value` is not controlled by the D
...
 
 ```jsx
 // Uncontrolled, because Preact doesn't set the value
 <input onInput={myEventHandler} />;
-
-// Controlled, because Preact manages the input's value now
-<input value={someValue} onInput={myEventHandler} />;
 ```
 
-> One gotcha to note here is that setting the value to `undefined` or `null` will essentially become uncontrolled.
```
rschristian, line 32: Not sure that this is particularly relevant given our newer controlled example.

JoviDeCroock (approved):

PR #1174, opened 2024-08-07
blog: Prerendering with `@preact/preset-vite`

rschristian:
Not much of a writer so let me know if there are any opinions on topics, direction, etc.!

### DIFF
--- a/content/en/blog/prerendering-preset-vite.md
+++ b/content/en/blog/prerendering-preset-vite.md
@@ -290,6 +290,7 @@ export async function prerender(data) {
 		initialized = true;
     }
     if (!initialized) await init();
+    ...
 }
 ```
 
--- a/content/en/blog/prerendering-preset-vite.md
+++ b/content/en/blog/prerendering-preset-vite.md
@@ -7,9 +7,9 @@ authors:
 
 # Prerendering with Preset Vite
 
-It's been a half-year since our prerendering plugin has somewhat quietly become available in `@preact/preset-vite`, so let's talk about it, our history, and the ecosystem at large
+It's been a half-year since our prerendering plugin has somewhat quietly become available in `@preact/preset-vite`, so let's talk about it, our history, and the ecosystem at large.
 
-Those who have been in our community for a while know how much we like prerendering; it was a first-class feature in Preact-CLI, then WMR, and now our Vite preset. When it's done right, it's a pain-free addition to the typical SPA that improves the user experience greatly and this plugin aims to facilitate just that.
+Those who have been in our community for a while know how much we like prerendering; it was a first-class feature in Preact-CLI, then WMR, and now our Vite preset. When it's done right, it's a pain-free addition to the typical SPA that improves the user experience greatly and our prerender-plugin aims to facilitate just that.
 
 ## What is Prerendering?
 
--- a/content/en/blog/prerendering-preset-vite.md
+++ b/content/en/blog/prerendering-preset-vite.md
@@ -341,8 +341,8 @@ The patch in question is as follows:
 	 ...
 ```
 
-As attempting to execute `document.getElementsByTagName` will error in Node where there is no `document` we simply add an additional condition to the preloader so that it makes no attempt to run in Node and that's it. Just the partial change of this one line.
+As attempting to execute `document.getElementsByTagName` will error in Node where there is no `document`, we simply add an additional condition to the preloader so that it makes no attempt to run in Node, and that's it. Just the partial change of this one line.
 
-However, we are very, very happy with this level of risk and have been suing it heavily for some time now without any issue, but, we are somewhat using the tool beyond what it was intended for and it's something we want to disclose.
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

content/en/blog/prerendering-preset-vite.md
```diff
@@ -0,0 +1,348 @@
...
+date: 2024-08-06
+authors:
+  - Ryan Christian
+---
+
+# Prerendering with Preset Vite
+
+It's been a half-year since our prerendering plugin has somewhat quietly become available in `@preact/preset-vite`, so let's talk about it, our history, and the ecosystem at large
+
+Those who have been in our community for a while know how much we like prerendering; it was a first-class feature in Preact-CLI, then WMR, and now our Vite preset. When it's done right, it's a pain-free addition to the typical SPA that improves the user experience greatly and this plugin aims to facilitate just that.
```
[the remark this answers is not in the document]
rschristian: Thanks!

content/en/blog/prerendering-preset-vite.md
```diff
@@ -0,0 +1,348 @@
...
+
+```diff
+// src/node/plugins/importAnalysisBuild.ts
+-if (__VITE_IS_MODERN__ && deps && deps.length > 0) {,
++if (__VITE_IS_MODERN__ && deps && deps.length > 0 && typeof window !== 'undefined') {,
+	 const links = document.getElementsByTagName('link')
+	 ...
+```
+
+As attempting to execute `document.getElementsByTagName` will error in Node where there is no `document` we simply add an additional condition to the preloader so that it makes no attempt to run in Node and that's it. Just the partial change of this one line.
```
rschristian, line 344:
```suggestion
As attempting to execute `document.getElementsByTagName` will error in Node where there is no `document`, we simply add an additional condition to the preloader so that it makes no attempt to run in Node, and that's it. Just the partial change of this one line.
```

content/en/blog/prerendering-preset-vite.md
```diff
@@ -0,0 +1,348 @@
...
+
+However, we are very, very happy with this level of risk and have been suing it heavily for some time now without any issue, but, we are somewhat using the tool beyond what it was intended for and it's something we want to disclose.
```
rschristian, line 346:
```suggestion
We are very, very happy with this level of risk and have been using it heavily for some time now without any issue, but, this is somewhat using the tool beyond what it was intended for and it's something we want to disclose.
```

content/en/blog/prerendering-preset-vite.md
```diff
@@ -0,0 +1,351 @@
...
+	- Some amount of buy-in/lock-in
+	- Support matrix for different Vite config options, plugins, etc. can be complicated and less than clear
+
+While these solutions absolutely have their merits and places in the ecosystem, neither felt as great as they could be for our ecosystem, given our historic offerings in this area. The "best case scenario" DX was often sacrificed for more complex or specific needs -- which is a completely valid trade off.
+
+For drop-in prerendering, however, we thought we could provide something a bit different to the existing options, or at least something a bit more familiar to our users.
+
+## Implementation in `@preact/preset-vite`
+
+Year later, we were still quite in love with the simplicity and extensibility of WMR's prerendering and felt it was sorely missing from our Vite preset, so we looked to port it over with a few minor adjustment to fix the qualms we had. A little bit of work later and voila, prerendering via a plugin!
```
marvinhagemeister, line 46: Should this be `A year later,...`?

rschristian: `Years`, thanks!

```suggestion
Years later, we were still quite in love with the simplicity and extensibility of WMR's prerendering and felt it was sorely missing from our Vite preset, so we looked to port it over with a few minor adjustment to fix the qualms we had. A little bit of work later and voila, prerendering via a plugin!
```

content/en/blog/prerendering-preset-vite.md
```diff
@@ -0,0 +1,351 @@
...
+if (typeof window !== 'undefined') {
+	hydrate(<App />, document.getElementById('app'));
+}
+
+export async function prerender(data) {
+    return await ssr(<App {...data} />)
+}
+```
+
+### Using `globalThis` to pass data around
```
marvinhagemeister, line 205: I like the later example of having a global cache to write to, but I'm a bit unsure about passing other data. I'm worried that this could lead to race conditions given that we also support async rendering in RTS now.

marvinhagemeister (commented):
I really like this blog posts. It's very well written. Left two minor comments

content/en/blog/prerendering-preset-vite.md
rschristian: There might be a better example in there. This sort of thing is something I've pointed out to people enough times that I think I want *something* there, as it's fairly reasonable to need to pass data back up (and out of the tree). Will have a think.

Shouldn't be any race conditions AFAIK; component tree will have to have fully rendered before the RTS pass completes? That's not to say you can't cause issues, but... shouldn't be anything inherent to the pattern?

marvinhagemeister (approved):

PR #1177, opened 2024-08-20
Update URL for Material UI Preact

ShiningTrapez:
Looks like this folder was renamed to add "UI" at some point, I've updated the URL accordingly


ShiningTrapez:
Yes, I can do that

rschristian:
Thanks!!

PR #1183, opened 2024-10-24
refactor: Move REPL query safety check into runner

rschristian:
I've been linking to a few REPL examples over the past few days, and one thing that's slightly bothering me is that we show the confirmation/warning for running untrusted code _before_ showing the code itself. The user has to gauge whether they trust whoever provided the link rather than if they trust the code (or copy the query into a base64 decoder and read that, I suppose).

This isn't perfect (see the `TODO:` I left behind), but it moves the confirmation check further ahead into the REPL runner itself. This lets the user (partially) see the code populated in the editor pane before they're forced to accept or refuse.

Before             |  After
before  | after

### DIFF
--- a/src/components/controllers/repl/runner.jsx
+++ b/src/components/controllers/repl/runner.jsx
@@ -136,8 +136,8 @@ export default class Runner extends Component {
 
 	async execute(transpiled, isFallback) {
 		// TODO: This is much better implemented as a dialog within the frame.
-		// `confirm` unforutnately is a window-level modal, so if there the
-		// input code is especially long, it could be hidden outside the page.
+		// `confirm` unforutnately is a window-level modal, so when/if the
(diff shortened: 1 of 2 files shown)


src/components/controllers/repl/runner.jsx
```diff
@@ -132,6 +135,24 @@ export default class Runner extends Component {
 	}
 
 	async execute(transpiled, isFallback) {
+		// TODO: This is much better implemented as a dialog within the frame.
+		// `confirm` unforutnately is a window-level modal, so if there the
+		// input code is especially long, it could be hidden outside the page.
```
rschristian, line 140:
```suggestion
		// `confirm` unforutnately is a window-level modal, so when/if the
		// code extends down past the window edge, the user can't see it.
```

JoviDeCroock (approved):

PR #1184, opened 2024-10-24
Update index.jsx

Manisha-95:
Added new company name which is using preact

### DIFF
--- a/src/components/we-are-using/index.jsx
+++ b/src/components/we-are-using/index.jsx
@@ -277,7 +277,13 @@ const companies = [
 		name: 'Loveholidays',
 		href: 'https://loveholidays.com',
 		logo: 'loveholidays.svg'
-	}
+	},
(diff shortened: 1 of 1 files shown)


src/components/we-are-using/index.jsx
```diff
@@ -277,7 +277,13 @@ const companies = [
 		name: 'Loveholidays',
 		href: 'https://loveholidays.com',
 		logo: 'loveholidays.svg'
-	}
+	},
+	{
+		name: 'Oracle',
+		href: 'https://www.oracle.com',
+		logo: 'Oracle.svg'
```
rschristian, line 284: You haven't added the SVG to the assets

rschristian (commented):
Do you work for Oracle? Generally we'd like only the companies themselves to suggest addition to this page.

PR #1186, opened 2024-10-29
feat: Basic content prefetch

rschristian:
Simple prefetching for content in a few cases:

- header links
- sidebar links
- links within markdown content
- blog overview links

Makes for a bit snappier nav on desktop.

### DIFF
--- a/src/lib/use-resource.js
+++ b/src/lib/use-resource.js
@@ -1,8 +1,17 @@
 import { useEffect } from 'preact/hooks';
 import { getContent } from './content.js';
 
-/** @type {Map<string, Promise<any>>} */
+/**
+ * @typedef {Object} CacheEntry
+ * @property {Promise<any>} promise
+ * @property {'pending'|'success'|'error'} status
+ * @property {any} result
+ * @property {number} users
+ */
+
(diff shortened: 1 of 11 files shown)


src/lib/use-resource.js
```diff
@@ -1,41 +1,46 @@
...
-/** @type {Map<string, Promise<any>>} */
+/** @type {Map<string, CacheEntry>} */
 const CACHE = new Map();
+const createCacheKey = (fn, deps) => '' + fn + JSON.stringify(deps);
+
+/**
+ * @param {string} path
+ */
+export function prefetchContent(path) {
+	const lang = document.documentElement.lang;
```
rschristian, line 22: This shouldn't be problematic, it's kept sync'd to the context value normally:

https://github.com/preactjs/preact-www/blob/67554f57a84e700ddeb435c3d654d6a78d3f068d/src/lib/i18n.jsx#L56-L59

While components could just pass it in & we can use the context normally, it's a bit more verbose and there's no reason to prefetch anything other than the currently set language, so abstracting it away is nicer.

JoviDeCroock (approved):

PR #1197, opened 2024-11-19
Add Browserify aliasing instructions

joepie91:
As the title implies :)

### DIFF
--- a/content/en/guide/v10/getting-started.md
+++ b/content/en/guide/v10/getting-started.md
@@ -207,6 +207,30 @@ module.exports = {
 };
 ```
 
+#### Aliasing in Browserify
+
+To alias within Browserify, you'll need to install [aliasify](https://www.npmjs.com/package/aliasify) first.
+
+Then, assuming you are passing the configuration to Browserify from within JS code, you'll have to add the following configuration *after* `babelify` (or any other language transpilers) in your Browserify configuration, like so:
+
+```js
+browserify(files, {
+  // ...
+  transform: [
+    // ... other transforms like babelify go here ...
+    ["aliasify", {
+      aliases: {
+        "react": "preact/compat",
+        "react-dom": "preact/compat"
+      }
+    }],
+    // ...
+  ]
+})
+```
+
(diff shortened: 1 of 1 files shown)


rschristian (commented):
Thanks, but given Browserify's super legacy status, I'm not sure we really need this. Probably seems a bit silly, but adding in legacy tools does alter perception a bit for the worse.

The `aliasify` module being without updates for 8+ years also makes it a rather concerning thing to recommend, even if it might be the de facto tool for the ecosystem.

joepie91:
It seems odd to brand Browserify as 'legacy' merely because it happens to be older software; it continues to be used in many places, given that the task it fulfills hasn't really changed.

Likewise I continue to use it, including in new projects; it is, for many cases, vastly simpler and more predictable to operate than newer bundlers. That's where this PR originates from - me setting it up for a new project where I'm trying out Preact (and I have not run into any issues with that, `aliasify` is still working as fine as it did 8 years ago).

I do have to admit that drawing this equivalence between "old" and "legacy" does not inspire much confidence for the long-term support of Preact, and this is not the response I'd expected...

rschristian:
"Legacy" not so much because it's old, but because it hasn't modernized. Front page is still suggesting CJS-authored web apps in 2024 ¯\\\_(ツ)\_/¯

It's great if you like & use that, I'm certainly not trying to convince you to use anything else! I don't think there's any reason, besides familiarity, that one would reach for it today. Heck, we even get some comments about why we continue to document _Webpack_. Right or wrong (and I am on the side of thinking it's a bit silly), which tools we present does alter perception, especially as a tool that was early to ESM and really, really strongly suggests it.

> I do have to admit that drawing this equivalence between "old" and "legacy" does not inspire much confidence for the long-term support of Preact

Considering Preact X still supports IE11, not sure where that comment is coming from. Here's even a blog post on our stability: https://preactjs.com/blog/preact-x

But this is OSS, you're absolutely free to go elsewhere if you wish.

---

Thanks, but I don't think we need to document Browserify these days.

PR #1202, opened 2024-12-04
refactor: Lazy load algolia docsearch

rschristian:
Per #1106, it looks like I was doubting the value of lazy loading this. Certainly possible I was wrong to do so though.

### DIFF
--- a/src/components/header/search.jsx
+++ b/src/components/header/search.jsx
@@ -1,4 +1,4 @@
-import { lazy } from 'preact-iso';
+import { lazy, ErrorBoundary } from 'preact-iso';
 import style from './style.module.css';
 import config from '../../config.json';
 
@@ -18,12 +18,14 @@ const transformItems = (items) =>
 export default function Search() {
 	return (
 		<div class={style.search}>
-			<DocSearch
-				apiKey={config.docsearch.apiKey}
-				indexName={config.docsearch.indexName}
-				appId={config.docsearch.appId}
-				transformItems={transformItems}
-			/>
+			<ErrorBoundary>
+				<DocSearch
+					apiKey={config.docsearch.apiKey}
+					indexName={config.docsearch.indexName}
+					appId={config.docsearch.appId}
+					transformItems={transformItems}
+				/>
+			</ErrorBoundary>
 		</div>
 	);
 }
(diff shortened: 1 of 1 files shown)


src/components/header/search.jsx
```diff
@@ -1,6 +1,8 @@
+import { lazy } from 'preact-iso';
 import style from './style.module.css';
 import config from '../../config.json';
-import { DocSearch } from '@docsearch/react';
+
+const DocSearch = lazy(() => import('@docsearch/react').then(m => m.DocSearch));
```
JoviDeCroock, line 5: Do we want a separate Suspenseful boundary for this so it doesn't bubble up to the route boundary?

rschristian: Could do, yeah.

I'm actually going to be away for the next few days (gotta get to bed now) so if you (or anyone else) wanted to add on top here certainly feel free. Else I'll hopefully get back to this in a few.

src/components/header/index.jsx
```diff
@@ -35,7 +36,9 @@ export default function Header() {
 			<div class={style.outer}>
 				<div class={style.inner}>
 					<Nav class={style.nav} routes={config.nav} current={url} />
-					<Search />
+					<Suspense fallback={<span>Search</span>}>
+						<Search />
+					</Suspense>
```
rschristian, line 41: Will already be using `/compat` as their search is a React component, I don't think a spoofed suspense boundary would do much good here?

As for the fallback, shouldn't ever be shown to my knowledge so probably doesn't matter. Just went with text to avoid too much shifting in dev & throttled connections more than anything.

src/components/header/search.jsx
```diff
@@ -16,12 +19,14 @@ const transformItems = (items) =>
 export default function Search() {
 	return (
 		<div class={style.search}>
-			<DocSearch
-				apiKey={config.docsearch.apiKey}
-				indexName={config.docsearch.indexName}
-				appId={config.docsearch.appId}
-				transformItems={transformItems}
-			/>
+			<Suspense fallback={<p>Search</p>}>
```
rschristian, line 22: Will already be using `/compat` as their search is a React component, I don't think a spoofed suspense boundary would do much good here?

As for the fallback, shouldn't ever be shown to my knowledge in prod so probably doesn't matter. Just went with text to avoid too much shifting in dev & throttled connections more than anything.

JoviDeCroock: Could use ErrorBounday from preact-iso as well if you prefer

JoviDeCroock (approved):

PR #1203, opened 2024-12-04
refactor: Fix lighthouse accessibility issue in footer links

rschristian:
Can see this from local lighthouse runs or [pagespeed insights](https://pagespeed.web.dev/analysis/https-preactjs-com/7ipoxsksca?form_factor=mobile).

temp

While the message seems to suggest a contrast issue, this isn't actually the case -- our contrast ratios are 6.53:1 & 6.01:1 w/ light & dark preferences respectively, both hitting the WCAG AA standards. In fact, changing this to #fff on #000 still does nothing to dismiss the message. The actual issue, which I found unclear at least, is that links aren't meant to rely on color alone and so we need text decoration here. Reasonable enough.

### DIFF
--- a/src/components/footer/style.module.css
+++ b/src/components/footer/style.module.css
@@ -12,7 +12,25 @@
 		margin-left: 0.625rem;
 	}
 
+	/**
+	 * Ideally kept sync'd with the markdown links:
+	 * https://github.com/preactjs/preact-www/blob/f68e65aadbd19b042921cc175aa256f3aa248768/src/style/markdown.css#L18
+   */
 	a {
 		color: var(--color-footer-link);
+		text-decoration-skip-ink: auto;
+		text-decoration: underline;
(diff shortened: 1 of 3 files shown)


src/components/footer/style.module.css
```diff
@@ -12,7 +12,25 @@
 		margin-left: 0.625rem;
 	}
 
+	/**
+	 * Ideally kept sync'd with the markdown links:
+	 * https://github.com/preactjs/preact-www/blob/f68e65aadbd19b042921cc175aa256f3aa248768/src/style/markdown.css#L18
+   */
```
rschristian, line 18: As this comment would suggest, our new footer link styles should match the markdown links but we're relying on copy/paste for the moment. It's only a few lines and I don't imagine it'll change often.

JoviDeCroock (approved):

PR #1205, opened 2024-12-07
refactor: Bump marked, move most HTML transforms into it

rschristian:
Not quite as clean as I'd like but it's an improvement.

Nearly all HTML transforms (like adding `target="_blank"` to external links or fragment links to headings) are now handled by `marked` at build time rather than `preact-markup` at runtime. Shouldn't be noticeable difference in terms of bundle or perf, but will be slightly better for the latter in theory.

HTML in markdown is the one hair in the soup though, as `marked` will not handle it. We only make use of it on the home page really, not sure if we can refactor that to avoid this. As such, had to update some of the `/index.md` documents to add `decoding="async"` to them, as `preact-markup` no longer will, and we'll need to keep around the HTML parser-based highlighting for those code blocks/demos at the very bottom of the home page.


content/de/about/browser-support.md
```diff
@@ -7,12 +7,5 @@ permalink: '/about/browser-support'
 
 Prect unterstützt moderne Browser (Chrome, Firefox, Safari, Edge) und IE9+. Es sollte in IE7 sowie IE8 funktionieren, setzt allerdings einige Polyfiller voraus. Wenn es beabsichtigt ist, Preact zu benutzen und diese älteren Browser zu unterstützen, wird empfohlen, einen Polyfiller wie zum Beispiel [es5-shim] oder [babel-polyfill] zu verwenden.
 
-<center>
-    <a href="https://saucelabs.com/u/preact">
-        <img src="https://saucelabs.com/browser-matrix/preact.svg" alt="Browserlist">
-    </a>
-</center>
```
rschristian, line 14: The saucelabs link is dead & has been for ages. Removed it

content/de/guide/v8/progressive-web-apps.md
```diff
@@ -49,7 +49,7 @@ Preact ist eine ausgezeichnete Wahl für [progressive Web Apps](https://web.dev/
 
 In der Tat ist genau diese Website eine Progressive Web App! Hier ist zu sehen, wie die App am Beispiel einer Trace ausgehend von einem Nexus 5X über 3G in unter 5 Sekunden interaktiv wird:
 
-<img src="/pwa-guide/timeline.jpg" alt="Eine DevTools-Zeitleisten-Trace der preactjs.com-Seite auf einem Nexus 5X"/>
+![Eine DevTools-Zeitleisten-Trace der preactjs.com-Seite auf einem Nexus 5X](/pwa-guide/timeline.jpg)
```
rschristian, line 52: Simply an unnecessary use of HTML when normal markdown would do & is preferable

content/de/index.md
```diff
@@ -40,7 +40,7 @@ function Counter() {
 
 
 <section class="home-section">
-  <img src="/home/metal.svg" alt="metal">
+  <img src="/home/metal.svg" alt="metal" loading="lazy" decoding="async" width="54" height="54">
```
rschristian, line 43: `decoding="async"` is the only real addition here, lazy & the sizes have been added to the English doc ages ago but hasn't quite made it to all the translations. Figured I'd fix it now.

plugins/precompile-markdown/index.js
```diff
@@ -103,26 +153,23 @@ function generateToc(markdown) {
...
 	}
 	return toc;
 }
 
-/*
- * Many markdown formatters can generate the table of contents
- * automatically. To skip a specific heading the use an html
- * comment at the end of it. Example:
- *
- *   ## Some random title <!-- omit in toc -->
```
rschristian, line 122: We've never made use of this, and I don't imagine we will anytime soon

src/lib/marked.worker.js
```diff
@@ -1,10 +0,0 @@
-import marked from 'marked';
```
rschristian, line 1: Leftover, long-since unused.

marvinhagemeister (approved):
This is really great! I'm a big fan of doing those transformations as early as possible in the pipeline. Great work!

PR #1206, opened 2024-12-07
refactor: Add bash syntax highlighting

rschristian:
It's our only missing highlight language, and now that Prism is server-only, I figure it's worth adding now (as it's a bit chonky).

The "Getting Started" & "Unit Testing with Enzyme" pages have a few examples to look at.


marvinhagemeister (approved):

marvinhagemeister:
Since we're doing prism server-side only now could also switch to importing the languages from the `prismjs` package in the future.

Issue #1001, opened 2023-06-14
Callback Refs note on double call is confusing


L1Q:
On the Refs guide page, there is a note:

> If the ref callback is defined as an inline function it will be called twice. Once with `null` and then with the actual reference. This is a common error and the `createRef` API makes this a little easier by forcing user to check if `ref.current` is defined.

[🔗The relevant source file](https://github.com/preactjs/preact-www/blob/master/content/en/guide/v10/refs.md?plain=1#LL67C1-L68C1)
[🔗The article](https://preactjs.com/guide/v10/refs#callback-refs)

<details>
  <summary>Screenshot of the piece in question</summary>

![image](https://github.com/preactjs/preact-www/assets/6942070/1b6ce965-b888-4681-865b-06d0ba36adf0)
</details>

This note is useful in the sense that it's correct, in the first part that is:
``If the ref callback is defined as an inline function it will be called twice. Once with `null` and then with the actual reference.``
Indeed the ref function will be called twice, e.g. during a Vite's HMR update.

Where it gets crazy is the second half:
``This is a common error``
Very ambiguous here. On whose behalf is the error? What is the actual issue? Do you keep reintroducing the "error" in every preact release?

Now after some looking around I figured, perhaps there is a good reason it's called twice?
From what I found, it looks like the first call ever is always non-null, the second call will pass null, immediately followed by the third call, again with an actual element passed.

It looks like null is passed during unmount, and element is passed during mount.

Is this a correct assumption? Can this be clarified in the guide?

rschristian:
> On whose behalf is the error? What is the actual issue? Do you keep reintroducing the "error" in every preact release?

It's an issue for the users, as it might be unexpected. It's not an "error" in Preact releases, but fundamentally how the API works.

> Now after some looking around I figured, perhaps there is a good reason it's called twice?

If you define the function inline, every render will create a new instance of this function. Because of this, the function will be called once with `null` to clear it out then again with the actual DOM element to set the ref back up.

Indeed, expounding upon that might be good. Changing that from "common error" to "common issue" is probably a good idea too.

L1Q:
Funny you mention ``inline`` specifically. Perhaps my understanding of this term is different from yours.

In my testing, it didn't seem to matter whether or not the function was an arrow function defined directly inside the ``ref={}`` block. Defining it globally using a function keyword and then passing it to the ref prop produced the same behaviour.

rschristian:
> Funny you mention `inline` specifically.

[...]

PR #1212, opened 2024-12-24
refactor: Rewrite Refs doc

rschristian:
I've found the refs doc rather lacking for a while, hope this is an improvement. Notably, it adds hooks examples, mentions values other than DOM nodes, and mentions the new cleanup functions added in 10.23.

Should address #1001 too, I expanded the warning for the callback refs section to (hopefully) clarify that a bit better.

Edit: Will try to do something similar to the Context doc, either during more gaps in holiday celebrations or in the next couple of weeks.

### DIFF
--- a/content/en/guide/v10/refs.md
+++ b/content/en/guide/v10/refs.md
@@ -1,6 +1,6 @@
 ---
 name: References
-description: 'References are a way of creating stable values that are local to a component instance but exist outside of the normal component lifecycle.'
+description: 'Refs are a way of creating stable values that are local to a component instance and persist across renders.'
 ---
 
 # References
--- a/content/en/guide/v10/refs.md
+++ b/content/en/guide/v10/refs.md
@@ -1,13 +1,15 @@
 ---
 name: References
-description: 'References can be used to access raw DOM nodes that Preact has rendered'
+description: 'References are a way of creating stable values that are local to a component instance but exist outside of the normal component lifecycle.'
 ---
 
 # References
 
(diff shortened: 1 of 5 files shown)


content/en/guide/v10/refs.md
```diff
@@ -1,124 +1,215 @@
...
+// Function components
+import { useRef } from "preact/hooks";
+
+function MyComponent() {
+  const countRef = useRef();
+  const inputRef = useRef(null);
+
+  // ...
+}
+```
```
rschristian, line 44: I'm thinking of adding a toggle widget to switch between class examples & hooks, might make this a bit cleaner.

Still have to think of a way to do that which doesn't break translated pages that might be out of sync though.

src/components/tab-group/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+ * so it's not necessarily meant for directly comparing API styles.
+ *
+ * Primarily adopted from MDN's tabs example:
+ * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example
+ *
+ * @param {Object} props
+ * @param {string} props.tabstring - Comma-separated list of tab names
+ * @param {Array<import('preact').Component | string>} props.children - Array of JSX elements
+ */
+export default function TabGroup({ tabstring, children }) {
```
rschristian, line 17: I'm super pleased with this (almost unreasonably so), should be great for showing off different API styles/options in the future.

As noted, MDN has a brilliant tutorial in which they go over all of the accessibility complexities which I believe I mirrored correctly. Everything seems to be labeled, connected, and following practices as intended.

JoviDeCroock (approved):
This is awesome

PR #1216, opened 2025-01-01
add some japanese translations for the nav

developit:
I've been learning Japanese and knew a few of these, other words I cross-checked with the Vue and Electron doc sites.


JoviDeCroock (approved):
Nice! Indentation looks a bit off

PR #1218, opened 2025-01-02
refactor: Rewrite Forms doc

rschristian:
Follow up to the refs & context rewrites, trying to add clarity here too.

Of note:
 - Adds hooks examples
 - Shows forms via `FormData`
 - Adds a suggestion to not rely on component state for all inputs
 - Tries to expound the Controlled Component issue

### DIFF
--- a/content/en/guide/v10/forms.md
+++ b/content/en/guide/v10/forms.md
@@ -1,13 +1,11 @@
 ---
 name: Forms
-description: 'How to build awesome forms in Preact that work anywhere.'
+description: 'Forms and form controls allow you to collect user input in your application and is a fundamental building block of most web applications.'
 ---
 
 # Forms
 
-Forms in Preact work much the same as they do in HTML. You render a control, and attach an event listener to it.
-
-The main difference is that in most cases the `value` is not controlled by the DOM node, but by Preact.
+Forms in Preact work in the same way as they do in HTML & JS: you render controls, attach event listeners, and submit information.
 
 ---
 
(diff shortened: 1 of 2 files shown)


content/en/guide/v10/forms.md
```diff
@@ -117,61 +176,241 @@ class MySelect extends Component {
...
+// --repl-after
+render(<MySelect />, document.getElementById("app"));
+```
+
+</tab-group>
 
-Checkboxes and radio buttons (`<input type="checkbox|radio">`) can initially cause confusion when building controlled forms. This is because in an uncontrolled environment, we would typically allow the browser to "toggle" or "check" a checkbox or radio button for us, listening for a change event and reacting to the new value.  However, this technique does not transition well into a world view where the UI is always updated automatically in response to state and prop changes.
+## Basic Forms
 
-> **Walk-Through:** Say we listen for a "change" event on a checkbox, which is fired when the checkbox is checked or unchecked by the user.  In our change event handler, we set a value in `state` to the new value received from the checkbox.  Doing so will trigger a re-render of our component, which will re-assign the value of the checkbox to the value from state.  This is unnecessary, because we just asked the DOM for a value but then told it to render again with whatever value we wanted.
```
rschristian, line 146: https://preactjs.com/guide/v10/forms#checkboxes--radio-buttons

I've read this through a dozen or so times and I'm still not quite sure what it is suggesting here. Using the updated state value from `onChange` & rerendering based upon it is silly & wasteful, so instead, use the opposite of the current value (as it's just a toggle) and rerender with that instead? It's the same issue, just a different format (and one I'd argue is more problematic, as it allows component state to get out-of-sync with the DOM in some cases).

Not sure if that makes sense to anyone else, very, very possible I'm misunderstanding entirely, but I stripped this comment out as it seemingly makes little sense to me at the moment. I don't understand what it's advocating for or why that's an improvement.

content/en/guide/v10/forms.md
```diff
@@ -117,61 +176,241 @@ class MySelect extends Component {
...
+const onInput = (e) => {
+  if (e.currentTarget.value.length <= 3) {
+    setValue(e.currentTarget.value);
+  }
+}
+```
+
+The problem with this is in the cases where the input fails that condition: because we don't run `setValue`, the component doesn't rerender, and because the component doesn't rerender, the input value is not correctly controlled. However, even if we did add a `else { setValue(value) }` to that handler, Preact is smart enough to detect when the value hasn't changed and so it will not rerender the component. This leaves us with [`refs`](/guide/v10/refs) to bridge the gap between the DOM state and Preact's state.
+
+> For more information on controlled components in Preact, see [Controlled Inputs](https://www.jovidecroock.com/blog/controlled-inputs) by Jovi De Croock.
```
rschristian, line 339: Tried to expound more on the controlled input issue, rather than jumping straight into "use refs", which I wrote before. I hope it's better? Basically trying to condense & transplant Jovi's writing.

Feedback on the wording here very much appreciated, as this is a bit of a prickly issue that users may run into.

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #1220, opened 2025-01-03
Fix typos in French translation

deining:
This PR fixes a few typos I spotted in the French translation.

### DIFF
--- a/content/fr/guide/v8/differences-to-react.md
+++ b/content/fr/guide/v8/differences-to-react.md
@@ -17,7 +17,7 @@ La raison pour laquelle Preact n'essaye pas d'inclure chaque fonctionnalité de 
 
 ## Compatibilité de version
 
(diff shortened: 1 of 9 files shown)


rschristian (approved):
Do you happen to be using a bot? Generally little reason to be referring to the v8 docs these days, they're 5+ years old (and probably diverge from the English docs). We do also have a policy that contributors [only author in their native language](https://github.com/preactjs/preact-www/blob/master/CONTRIBUTING.md#writing-content) for most cases, helps keep the quality up and avoid problematic mistranslations.

Not that this PR seems to be problematic itself, but I do want to avoid situations in which contributors are setting bots (or chatgpt and the like) loose on the repo to "correct" translations.

deining:
> Do you happen to be using a bot?

No.

> Generally little reason to be referring to the v8 docs these days

I agree.

> Not that this PR seems to be problematic itself

Glad to hear that. This PR will brings minor improvements, but I agree its relevance is limited.

> but I do want to avoid situations in which contributors are setting bots (or chatgpt and the like) loose on the repo to "correct" translations.

No, this is not my intent at all and won't happen.

Looking forward seeing my PR merged.

rschristian:
Fair enough.

"Generally little reason to be referring to the v8 docs these days" was more of a question why you were really looking at these, users shouldn't really be referring to them so it tends to be a good indicator of bots/automatic translations.

PR #1221, opened 2025-01-07
fix: typo in zh signals.md

linghaoSu:
Fixed a typo: changed '称为' to '成为'.
Note: “称为” means “to be called or named,” while “成为” means “to become a part of something.”


rschristian (approved):
Thanks!

PR #1222, opened 2025-01-07
feat: Preload routes and repl

rschristian:
Using the recent addition of the `.preload()` method from `preact-iso@2.8.0`, this PR preloads our few split chunks on mouse over and also preloads the REPL Code Editor & Runner chunks. Depending on how quick you are at clicking, can be a pretty considerable speed up.

Also refactors `use-resource` & `use-content` a bit, adding a new `use-repl` as well. Should be a bit easier to follow now.

### DIFF
--- a/src/components/header/style.module.css
+++ b/src/components/header/style.module.css
@@ -470,7 +470,6 @@
 	height: 56px;
 	min-width: 80px;
 	overflow: visible;
-	background: var(--color-brand);
 	padding-right: 0.5rem;
 
 	@media (max-width: /* --header-mobile-breakpoint */ 50rem) {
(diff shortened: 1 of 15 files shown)


src/components/header/style.module.css
```diff
@@ -470,7 +470,6 @@
 	height: 56px;
 	min-width: 80px;
 	overflow: visible;
-	background: var(--color-brand);
```
rschristian, line 473: Sorry, drive-by fix: the height of the search bar is ever so slightly larger than the other header items, and, on mobile, it overlaps (and I mean by 1-2px) into the Ukraine banner above it. We don't need a background here though, it's already sitting on a brand-colored header, so removing this fixes that.

JoviDeCroock (approved):

PR #1223, opened 2025-01-12
refactor: Use standard lazy loading for repl runner & editor

rschristian:
Sort of a follow up ttot #1222, I realized our loading of `<CodeEditor>` & `<Runner>` should just be using standard lazy loading. IIRC, I was screwing around with making the runner as a whole into a worker, hence that approach, but that never panned out and we were left with this bit of an oddity.


src/lib/repl.js
```diff
@@ -0,0 +1,30 @@
...
+// so we need to manually do it here to support the tutorial which reads the runner ref
+const oldDiff = options.__b;
+options.__b = (vnode) => {
+	if (vnode.type === Runner && vnode.ref) {
+		vnode.props.ref = vnode.ref;
+		vnode.ref = null;
+	}
+
+	if (oldDiff) oldDiff(vnode);
+};
```
rschristian, line 17: Might be worth adding this to `preact-iso`? I didn't realize we weren't forwarding refs for `lazy` there already.

JoviDeCroock (approved):

PR #1228, opened 2025-02-03
docs: Add info on `useImperativeHandle`

rschristian:
Since Jason added me to the Algolia instance a few weeks ago I've been getting weekly "round-up" summaries on the searches our site receives and for the past 4 weeks straight we've had a handful of failed searches for `useImperativeHandle`.

As such, I added an entry & tiny example for `useImperativeHandle`, creating a new section in our hooks docs to house both it and `useRef` as it seems to me like they should be grouped together.


JoviDeCroock (approved):
Hero

PR #1232, opened 2025-02-07
Fix links to the Todo List (Signals) example

nicolasiensen:
In this commit, we are fixing the link that should take users to the Todo List (Signals) example:
https://preactjs.com/repl?example=todo-signals

The same link was fixed in three different locales: en, ru and zh.


rschristian (approved):
Cheers

PR #1234, opened 2025-02-12
docs: Warn about using Preact in mixed CJS/ESM codebase without bundler

nwalters512:
I proposed this in https://github.com/preactjs/preact/issues/4648 but didn't hear back from a maintainer one way or another. I'm hoping to have a discussion about how to document this, as this was a really frustrating thing to discover independently.

### DIFF
--- a/content/en/guide/v10/getting-started.md
+++ b/content/en/guide/v10/getting-started.md
@@ -169,6 +169,8 @@ be seen in NextJS. To fix this, we can use aliases directly in our `package.json
 }
 ```
 
+> **Warning**: Preact is not recommended for use in a mixed CJS/ESM codebase running directly in Node without a bundler, as there's a risk that Preact's singletons could be loaded twice. If your first-party code or any dependencies will try to load both the CJS and ESM versions of Preact, you must use a bundler to alias imports of Preact to always resolve to either the CJS or ESM versions.
+
 #### Aliasing in Parcel
 
 Parcel uses the standard `package.json` file to read configuration options under


rschristian:
As mentioned, I think a variation of this would belong on a "Common Errors" page, rather than in our "Getting Started" docs, but unfortunately we don't have a page like that quite yet.

Might be better to convert this into an issue, we'll address it with docs when we can.

Thanks

nwalters512:
I'm likely biased because I faced this issue, but this feels like critical information to convey at the point when someone is first figuring out how to configure Preact in their project. While I understand you may have good reasons that you _can't_ fix this, the fact that Preact functions differently than most other dependencies is relevant upfront, not just when an error has already been encountered.

rschristian:
Sorry, but we see this as a problem specific to runtimes and it is one _that does_ affect many other packages. The bar for putting a warning right on our 'Getting Started', page which essentially amounts to "Use at your own risk", is exceptionally high and we feel hasn't been met here. It'd also be fairly odd to add this years after the fact when Node hasn't made things any worse than they have been.

Warnings for usage errors don't really belong on the first pages someone is likely to see, as these tend to be just as much for marketing as they are informational. I'm sorry that was a dealbreaker for you and sorry you spent time on an integration that wasn't going to work, but this information belongs elsewhere in our opinion.

As mentioned, feel free to open an issue for tracking if you'd like, else, it's something we're aware of and will keep in mind for if we do add a 'Common Errors'-type page.

Thanks,

nwalters512:
https://github.com/preactjs/preact-www/issues/1236

PR #1248, opened 2025-03-12
fix: Correct translation for "React" in Spanish

EGAMAGZ:
It looks like it was translated by a translator because instead of keeping React, it was translated into "reaccionar" (reaction)


rschristian (approved):
Thanks! We discourage use of automatic translator today but this might've been done years ago.

PR #1254, opened 2025-04-03
docs: Fix HTM example in 'No build workflows'

rschristian:
It worked as it was, but this addresses two things:

1. Bit silly to use JSX within the render call for a `htm` example (oops)
2. Some users get a bit stuck with how to refer to components using `htm`, and whilst the example does show using components already (`<${Button}>`), I figure an additional sample is good as it can be easy to miss in the component return. As such, I removed the `// --repl-after` comment so that it shows up in the code block on the docs page.


marvinhagemeister (approved):
Good catch!

Issue #1256, opened 2025-04-09
Inconsistency in the site across different languages.


grafos-lts:
**Describe the feature you'd love to see**
Remove outdated languages from the “https://preactjs.com” site.

**Additional context (optional)**

The preactjs.com site detects the language and displays the content accordingly.

It seems that some languages are no longer updated, and the documentation is outdated.

I noticed this situation when I followed the link to preact-cli (from the “English” site) and it announced “unfortunately no longer sees active development!”.

From the site selector I switched to “english” and noticed a big difference. 

Could you remove the languages that were no longer updated? Nowadays there are many very good translation tools, accessible, fast and very effective, in fact the preact website has the option from the navigation bar! And it works wonders!

Thanks for all your hard work and for creating this great tool!

English site:

![Image](https://github.com/user-attachments/assets/c2420dfc-6b21-4f35-b4ff-94310cf24aa3)

Same but in Spanish:

![Image](https://github.com/user-attachments/assets/2c349665-05a3-4009-b02b-71edef5fae6a)

About 'preact-cli':

![Image](https://github.com/user-attachments/assets/3249428b-2d88-4df5-adbe-08d40eca1fcb)

You can reproduce this issue with this selector at the bottom of the page;

![Image](https://github.com/user-attachments/assets/96a0f8c0-ada9-4765-bf5c-0aaa3901b7bc)

rschristian:
> I noticed this situation when I followed the link to preact-cli (from the “English” site)

Do you mean "from the Spanish site"?

Translations are entirely community-contributed, I don't think any of the core team speaks much Spanish, and we have actually received updates to the Spanish translations in the last month (see #1248 & the in-progress #1249).

Whilst this absolutely is a legitimate issue, I unfortunately don't know how to address it at the moment. Automatic translation services are pretty iffy still, I don't think we'd want to push people to them (especially as contributors have spent hours donating time to translate the documents correctly) and it's quite hard to determine automatically when a translation has fallen behind the English source of truth to flag it in some form.

Open to suggestions though.

grafos-lts:
Hi! First of all, thank you for responding, and so quickly!

You're absolutely right about the significant effort behind this—I apologize if my request came across the wrong way.

Yes, I was referring to the Spanish version of the site (I haven’t seen the other translations). As you can see in the screenshots, the information is outdated compared to the English version, which is confusing.

[...]

PR #1258, opened 2025-04-10
feat: Warn users for unmaintained translations

rschristian:
Re: #1256 

Introduces a sticky warning banner that warns users who view unmaintained translations (right now, everything but English & Russian). This warning banner provides a link to the English version as well as the sources here on GitHub in case any are willing to help contribute.

### DIFF
--- a/src/components/controllers/doc-page.jsx
+++ b/src/components/controllers/doc-page.jsx
@@ -1,3 +1,4 @@
+import { useState, useEffect } from 'preact/hooks';
 import { useRoute } from 'preact-iso';
 import { useContent } from '../../lib/use-content';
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
The warning is a bit large on mobile making the page 60% warning

rschristian:
Oof, didn't check that. Sorry.

Do we want it to be sticky? Non-sticky I suppose would correct that, though that only catches people who see the top of the page

JoviDeCroock:
Another alternative would be to use summary/details to have it be expandable.

Like, unmaintained translation warning which expands to what it is now

rschristian:
Oh that's a good shout, I'll give that a try.

Thanks!

PR #1263, opened 2025-04-13
chore: Normalize markdown content files

rschristian:
Sorry in advance, it's a pretty big diff. If it's preferable I can split up the changes to the markdown, I realize that's a scary-big list of files.

We haven't yet explicitly defined the markdown frontmatter anywhere, in types or otherwise, so it's been unclear what is used where. As such, we have a number of keys that are set but never used (`permalink`, `name`) and are missing keys that really should always be set (`title`, `description`). In some cases we're duplicating work to get around this, such as reading the markdown content to extract a `title` when it hasn't been set. Build-time only thing, so not necessarily a problem, but not ideal.

This PR tries to formalize this, creating types for the meta data in the files as well as transformed metadata (includes table of contents, whether it's a fallback, etc.) and cutting down on unused frontmatter keys. Should be no functional changes, just some cleaning up & refactoring.


src/components/table-of-contents/index.jsx
```diff
@@ -5,35 +5,31 @@ import { useContext, useMemo, useRef } from 'preact/hooks';
...
+	), [toc]);
 
-	if (items.length === 0) return <div ref={ref} />;
+	// TODO: Should we throw an error? No good reason to have a toc
+	// on the page if there is nothing to populate it with.
+	if (items.length === 0) return null;
 
 	return (
-		<div ref={ref}>
-			<nav onFocus={this.open}>
```
rschristian, line 24: This hasn't been hooked up for many years at this point, looks like we used to have a collapsible table of contents?

content/de/guide/v8/api-reference.md
```diff
@@ -1,13 +1,12 @@
...
-permalink: '/guide/api-reference'
+title: API-Referenzierung
 ---
 
 # API-Referenzierung
 
 ---
 
-<div><toc></toc></div>
+<toc></toc>
```
rschristian, line 9: I have a comment explaining this in `plugins/precompile-markdown/index.js`, but the gist of it is that the CommonMark spec defines which precise HTML elments can be used as a starting point for an HTML Block, and of course, something custom like `<toc>` doesn't fit the bill.

Without the wrapping `<div>` (as it is a valid starting identifier), `marked` would wrap the content in `<p>` which is not ideal. So instead, I added a little custom renderer to support `<toc>` here -- not essential, but removes an entirely superfluous node that serves an unclear purpose for everyone not familiar with CommonMark (which is probably everyone).

src/config.json
```diff
@@ -56,6 +56,12 @@
 			"ru": "Продолжить чтение"
 		},
 		"tutorial": {
+			"begin": {
+				"en": "Begin Tutorial",
+				"kr": "튜토리얼 시작",
+				"ru": "Начать обучение",
+				"zh": "开始教程"
+			},
```
rschristian, line 64: This seemed like a better fit for our config than in the individual markdown files.

src/components/table-of-contents/index.jsx
marvinhagemeister: yeah iirc we played around with that at some point. Not sure if we ever shipped it though.

marvinhagemeister (approved):
LGTM, great cleanup! I think many settings of front-matter were leftovers from past refactors.

PR #1265, opened 2025-04-13
refactor: Extract content paths util

rschristian:
We have a couple of markdown docs that differ in naming to the URL path, namely `/` -> `index.md` && `/tutorial` -> `/tutorial/index.md`

As the diff shows, this was previously handled at point of use, so the header had to correct paths, route components had to correct the paths, etc., and in this case `<ContentRegion>` was missing it so some links would 404 (`/tutorial.json`/`tutorial.md` doesn't exist). Better to handle this internally to `prefetchContent`/`useContent` as we can consistently apply the remapping & keep it all in one place.

### DIFF
--- a/src/components/content-region/index.jsx
+++ b/src/components/content-region/index.jsx
@@ -15,7 +15,7 @@ const COMPONENTS = {
 			const url = new URL(props.href, location.origin);
 
 			const prefetchAndPreload = () => {
-				if (props.href.startsWith('/repl?code')) {
+				if (props.href.startsWith('/repl')) {
 					ReplPage.preload();
 					preloadRepl();
 				} else if (props.href.startsWith('/tutorial')) {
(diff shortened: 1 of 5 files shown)


src/components/content-region/index.jsx
```diff
@@ -15,7 +15,7 @@ const COMPONENTS = {
 			const url = new URL(props.href, location.origin);
 
 			const prefetchAndPreload = () => {
-				if (props.href.startsWith('/repl?code')) {
+				if (props.href.startsWith('/repl')) {
```
rschristian, line 18: Not quite sure why I had originally appended the query param like that -- I'm not sure if we directly link to the REPL anywhere in our guide, so it probably doesn't matter, but it's wholly unnecessary. The REPL can be preloaded w/ or w/out a link to a particular code example.

marvinhagemeister (approved):
LGTM

PR #1268, opened 2025-04-17
chore: Bump most deps

rschristian:
Does what it says on the tin

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,5 +22,3 @@ jobs:
           cache-dependency-path: '**/package-lock.json'
       - run: npm ci
       - run: npm run build
-        env:
-          NODE_OPTIONS: '--openssl-legacy-provider'
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "vite build",
     "lint": "eslint src test",
     "format": "prettier --write \"{src,test}/**/*.{css,js,json}\"",
-    "prepare": "husky install"
+    "prepare": "npx simple-git-hooks"
   },
   "eslintConfig": {
     "extends": "developit",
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -22,5 +22,3 @@ jobs:
           cache-dependency-path: '**/package-lock.json'
       - run: npm ci
       - run: npm run build
-        env:
-          NODE_OPTIONS: '--openssl-legacy-provider'
```
rschristian, line 26: This was for Preact-CLI / Webpack 4, which we obviously don't need anymore

package.json
```diff
@@ -12,7 +12,7 @@
     "build": "vite build",
     "lint": "eslint src test",
     "format": "prettier --write \"{src,test}/**/*.{css,js,json}\"",
-    "prepare": "husky install"
+    "prepare": "npx simple-git-hooks"
```
rschristian, line 15: Husky and simple-git-hooks are pretty comparable but the former requires an extra config directory in the project root, which I'm not a huge fan of. Though a switch was fitting.

package.json
```diff
@@ -44,52 +44,54 @@
     "useTabs": true,
     "tabWidth": 2
   },
-  "lint-staged": {
-    "{src,test}/**/*.{css,js,json}": [
+  "simple-git-hooks": {
+    "pre-commit": "npx nano-staged"
+  },
+  "nano-staged": {
+    "**/*.{css,js,jsx,json}": [
```
rschristian, line 51: The lint-staged config was pretty restricted, so we're probably going to see a lot more formatting corrections pop up in the future here, though I think that's alright.

package.json
```diff
@@ -44,52 +44,54 @@
...
       "prettier --write"
     ]
   },
   "devDependencies": {
-    "@babel/polyfill": "^7.7.0",
-    "@xmldom/xmldom": "^0.8.10",
```
rschristian, line 54: Moved down to `"dependencies"`, making `"devDependencies"` here dev-only stuff (not that the distinction really matters w/ build tooling but I sometimes find it helpful)

JoviDeCroock (approved):

PR #1269, opened 2025-04-17
refactor: Reduce number of REPL chunks

rschristian:
Combines the code editor, REPL runner, splitter, and error overlay components into a singular chunk as they're always used together. They're not touched often enough to benefit from separate caching and the waterfall ends up being not great with them as-is.

Prior to this PR, we'd have 3 chunks: the editor, runner, and combined splitter & overlay. After, we have just the 1.


src/components/widgets.js
```diff
@@ -1,13 +1,16 @@
...
-import BlogOverview from './blog-overview';
 import Sponsors from './sponsors';
-import WeAreUsing from './we-are-using';
-import Branding from './branding';
 import TabGroup from './tab-group';
+import Toc from './table-of-contents';
+import TodoList from './todo-list';
+
+const Branding = lazy(() => import('./branding'));
+const WeAreUsing = lazy(() => import('./we-are-using'));
```
rschristian, line 13: Diff here is a bit whacky as I alphabetized them, made it a bit easier to search for.

`Branding` & `WeAreUsing` are the only widgets that I feel are valuable to lazy load at the moment, the rest are visible on the homepage or used often enough where it'd be detrimental to split them out at the moment (largely due to their CSS then being split, which we could fix, but that'd need some more work). Both only exist on one page each and `WeAreUsing` in particular ends up being fairly big as it contains a big ol' list of company names & logo hrefs.

JoviDeCroock (approved):

PR #1271, opened 2025-04-18
docs: Add preact-iso docs

rschristian:
"Router" has been the #1 search term for our docsearch every week since I got access ~4 months ago, with "route" and "location" also usually appearing in the top 10. Right now "router" links to a single reference of `preact-router` we have in the v8 upgrading docs which I don't think is offering users much value or what they're looking for. As such, I feel that we should offer some information here to get users started with routing as it's clearly a sought-after topic.

I've created a new section in the sidebar to accommodate this, thinking we can perhaps add more of our libraries and tools in the future. For example, `prefresh`, `@preact/preset-vite`, etc. I figure it shouldn't be too hard to keep in-sync, it's just a copy/paste of the readme for now.

### DIFF
--- a/content/en/guide/v10/preact-iso.md
+++ b/content/en/guide/v10/preact-iso.md
@@ -133,7 +133,7 @@ This will match the following routes:
 
 ## API Docs
 
-### `LocationProvider`
+### LocationProvider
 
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
This is such a missing piece, thank you!

PR #1291, opened 2025-06-09
docs[tanslate]:Translate the untranslated Chinese parts.

SisyphusZheng:
The translation work covers two scopes: about and blog. 
Currently, the Chinese documentation is basically consistent with the English documentation.


rschristian (approved):
Thank you so much! We really appreciate it.

PR #1292, opened 2025-06-10
refactor: Fix header styling issues

rschristian:
Will review & explain the changes one-by-one below, but looks to fix a handful of issues that have crept up. None are too problematic but good opportunity to review & fix some things.


src/components/header/style.module.css
```diff
@@ -255,12 +255,6 @@
 			& > button {
 				background: #fff;
 				color: #555;
-				z-index: 101;
-
-				svg {
-					filter: invert(40%) sepia(0%) saturate(3773%) hue-rotate(315deg)
-						brightness(73%) contrast(81%);
-				}
```
rschristian, line 263: The `z-index` is entirely unused, goes back to the repo creation. This style would only apply to the 'About' or i18n dropdown buttons which wouldn't need a z-index as they're kept inline, exactly where they were.

The SVG filter is a far more complex method of updating the icon color; way easier to just switch the icon to use `currentColor` and piggyback off the color setting just a few lines up (`color: #555`).

src/components/sidebar/style.module.css
```diff
@@ -21,7 +21,7 @@
 	color: #fff;
 	fill: rgba(255, 255, 255, 0.6);
 	cursor: pointer;
-	z-index: 560;
+	z-index: 460;
```
rschristian, line 24: I think this was a typo.

The slide-out sidebar has a `z-index` of `450`, and the slide-down header has a `z-index` of `500`. As such, when the header is expanded, there's really no reason for the sidebar toggle button to be displayed -- you're toggling something that can't be seen.

`z-index` of `460` allows it to still sit on top of the sidebar when expanded but hidden behind the header if it is expanded.

src/components/header/style.module.css
```diff
@@ -29,7 +31,7 @@
 	.open & {
 		justify-content: space-between;
 		-webkit-overflow-scrolling: touch;
-		height: var(--vh);
+		height: calc(var(--vh) - var(--banner-height));
```
rschristian, line 34: Predates my involvement, but IIUC, this was a problem introduce when the banner was added. It seems like the idea was that the social icons would be at the bottom of the screen when the togglable header was expanded, however, it never took into account the banner height. As such, the icons were always clipped off the screen.

The handful of white pixels at the very bottom of the screen are our icons, but you obviously can't interact with them as-is.

Screenshot of the site on a mobile device with the header expanded. Shows just the very top of icons (a handful of pixels at most)

src/components/header/style.module.css
```diff
@@ -94,11 +96,7 @@
 	}
 
 	nav {
-		position: relative;
-		display: inline-block;
-		font-size: 100%;
-		user-select: none;
-		-webkit-user-drag: none;
```
rschristian, line 101: AFAICT, completely unnecessary.

src/components/header/style.module.css
```diff
@@ -94,11 +96,7 @@
...
+		display: flex;
```
rschristian, line 99: Fixes the issue of header items wrapping on some screensizes:

Screenshot of the site at a width of 900px. The header nav items wrap so that the 'REPL' link is pushed out of the header and on top of the page content

Still not perfect, there's now some overlapping in the header, but wrapping is very bad.

src/components/header/style.module.css
```diff
@@ -185,6 +183,8 @@
 
 		& > a,
 		& > button {
+			display: flex;
+			align-items: center;
```
rschristian, line 187: Ensures the little dropdown arrow next to 'About' (or any future dropdown pages) is vertically centered:

Screenshot of the site header. There's a triangular dropdown arrow next to the 'About' link, but it's a bit lower than center compared to the label

src/components/header/style.module.css
```diff
@@ -313,6 +307,8 @@
 	justify-content: space-between;
 	align-items: flex-start;
 
+	touch-action: none;
+
```
rschristian, line 311: When the header is opened, we should not let the user scroll the page in the background.

src/components/header/style.module.css
```diff
@@ -327,21 +323,38 @@
 .translation {
 	position: absolute;
 	height: 2rem;
-	top: calc(var(--vh) - 3rem); /* TODO: Why does bottom not work? */
-	display: flex;
+	display: none;
+	bottom: calc(env(safe-area-inset-bottom) + 1rem);
```
rschristian, line 327: Hides the icons by default and uses the more straightforward `bottom` now that our parent height has been corrected (see comment above).

What it looks like now:

Screenshot of header expanded on mobile. The social icons now sit within the viewport at the bottom of the screen

src/components/header/style.module.css
```diff
@@ -399,7 +412,6 @@
 	padding: 18px 23px 12px;
 	line-height: var(--header-height);
 	text-align: center;
-	z-index: 100;
```
rschristian, line 402: Superfluous `z-index`, it sits on the header (`z-index` of `500` already) and nothing overlaps it either.

marvinhagemeister (approved):
All great fixes! Quality work as usual 👍

PR #1306, opened 2025-07-14
docs: Add compat-only hooks, `useDebugValue`, and expound upon `useLayoutEffect`

rschristian:
We're missing a number of compat-only hooks from our docs and per our Algolia stats, a number of user are searching for them. As such, I felt it was worth adding docs even for the hooks that are stubbed out.

While I was at it I noticed we were missing docs for `useDebugValue` and our `useLayoutEffect` description was quite light, so I added them too.


plugins/generate-llms-txt.js
```diff
@@ -83,6 +83,7 @@ export default function generateLlmsTxtPlugin(options = {}) {
 
 	return {
 		name: 'generate-llms-txt',
+		apply: 'build',
 		async buildStart() {
```
rschristian, line 87: I mentioned this in the other PR but looks like it was missed: `this.emitFile` cannot be used in Vite's dev mode and it'll emit warnings over this. The closest equivalent for us would be adding a middleware likely, as nothing else consumes the result, but in lieu of that, we should limit this to builds.

content/en/guide/v10/hooks.md
```diff
@@ -514,3 +535,98 @@ const App = props => {
...
+
+```jsx
+import { useDebugValue, useState } from 'preact/hooks';
+
+function useCount() {
+	const [count, setCount] = useState(0);
+	useDebugValue(count, c => `Count: ${c}`);
+	return [count, setCount];
+}
+```
```
rschristian, line 565: Side note: the formatter arg is lazy in React, hence why it exists, but our impl is very much not. It immediately calls the formatter upon ever render of the component, regardless of whether devtools are inspecting the element.

Not sure if that was intentional or not.

JoviDeCroock (approved):

PR #1311, opened 2025-07-25
fix: todo list style fixed

rajaboff-dev:
https://preactjs.com/ you can see here that if more todos are added to the todo list, its height keeps increasing, which I think is a mistake.

### DIFF
--- a/src/style/home.css
+++ b/src/style/home.css
@@ -189,5 +189,9 @@ main .markup {
 			box-sizing: border-box;
 			width: 75%;
 		}
+		ul {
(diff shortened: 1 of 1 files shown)


rschristian:
Thanks, but I think this is fine? If a user wants to input a dozen+ todo elements, it's not really an issue if the UI expands to accommodate, IMO.

rschristian:
Yeah, I see what you've tried to do but it doesn't really work on mobile and I don't think it's an issue regardless really. Keeping it as-is is fine.

Thanks though!

PR #1317, opened 2025-07-31
refactor: Rebuild header components

rschristian:
Forewarning, sorry, this is fairly big.

Refactors most of the header to make everything a bit more fit for purpose but there are no real functional changes. It corrects a few (IMO) minor grievances that have popped up whilst making other edits:

- Our usage / needs have drifted over the years compared to when most of components were built. Plenty of this is my doing, but for example:
  - Whilst we had a `<NavMenu>` for the "About" routes, we later added a language drop down. This overlapped somewhat with the drop-down menu, but not completely, so there's this [awkward ternary](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L160-L182) in the component.
  - We needed to list `/branding` in `config.nav` so that it was detected as a valid route (404 detection and whatnot), but that list makes up the header items, so we [had to add a magic prop](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L194)
  - The "About" menu is sorta a link, sorta not, so we add [an additional conditional](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L196-L207) to give it special handling too. As such, sometimes `<NavLink>` returns an actual link, sometimes it's just a button that triggers no navigation whatsoever. A bit weird.
  - We have a [`pathMatchesRoute` function](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L290-L304) to apply the "selected" header style, but then have added [additional conditionals](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L116-L117) outside of that because the function doesn't actually cover the cases we need.
  - TLDR: We've quickly appended stuff over the years. Happens in every project, but this should hopefully be a bit better for us & easier to adjust in the future when needed.
- Starts to break up `config.json`'s responsibilities
  - Our `config.json` is a bit of a weird god-file, in that it controls templating (`config.nav` -> header links), route definitions (used in `/lib/route-utils.js` for 404 detection & prerendering), and internationalization, all at once.
  - As such, if you wanted to add a class name to the "Home" link in the header, [you're doing it via JSON properties](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/config.json#L83), not the JSX. Until you need to use variables/components, that is, then you're left [matching routes](https://github.com/preactjs/preact-www/blob/0d52a6d9397f4ad9dda77b48c6b6f8a16690816d/src/components/header/index.jsx#L215-L217).
- Language picker is inaccessible
  - My bad, the picker items are spans missing tabindexes, not that tabindexes there would be preferable over the alternatives anyhow. Can't select language via tabbing through the doc.
  - Swapped the spans for buttons, easy fix at least. Dunno what I was thinking there.

The responsibilities of `config.json` is still a problem that still needs work, but this has at least gotten the templating out for the most part -- no where else are we really passing props via JSON. Ideally, I'll get routes split out from the internationalization in a sane way and then write a plugin to split the i18n file per language, should be about ~5kb of savings there. You'll only need 1 locale at a time of course.

This ships a tiny bit less JS, but nothing stellar. ~1kb less.

### DIFF
--- a/src/lib/i18n.jsx
+++ b/src/lib/i18n.jsx
@@ -1,6 +1,7 @@
 import { createContext } from 'preact';
 import { useContext, useEffect, useState } from 'preact/hooks';
 import { useLocation } from 'preact-iso';
+
 import { localStorageGet, localStorageSet } from './localstorage';
 import config from '../config.json';
 
@@ -49,7 +50,7 @@ export function LanguageProvider({ children }) {
 		document.documentElement.lang = userLang;
 	}, []);
 
-	const setAndUpdateHtmlAttr = (lang) => {
+	const setAndUpdateHtmlAttr = lang => {
 		localStorageSet('lang', lang);
 		setLang(lang);
 		document.documentElement.lang = lang;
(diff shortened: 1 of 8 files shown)


src/lib/i18n.jsx
```diff
@@ -72,11 +73,44 @@ export function useLanguage() {
...
+/**
+ * @param {{ name: Record<string, string> | string }} route
+ * @param {string} lang
+ * @return {string}
+ */
+export function getRouteName(route, lang) {
+	return typeof route.name === 'object'
+		? route.name[lang] || route.name.en
+		: route.name;
+}
```
rschristian, line 116: More of a hold over mechanism, I hope to switch the nav translations into a keyed object & break them away from the route paths in the future.

PR #1321, opened 2025-08-13
docs: Swap out "GitHub Stars" homepage demo for "GitHub Repos"

rschristian:
The 'GitHub Stars' example we have on the bottom of the home page is not great, IMO, as it's calling an external & hidden function in `componentDidMount`. Doesn't show much beyond that you can update state after mount to display some data, and even if we moved in the fetch call, it's a bit plain.

This PR swaps that out with a slightly altered version of our 'GitHub Repos' example from the REPL which showcases a similar situation (fetching external data & consuming it) but does so transparently and slightly more interestingly.


src/lambda/release.js
```diff
@@ -11,7 +11,7 @@ export default async function releaseLambda(req, _context) {
 		status: 200,
 		headers: {
 			'Content-Type': 'application/json',
-			'Cache-Control': 'public, max-age=3600, stale-if-error=86400'
+			'Cache-Control': 'public, max-age=86400, stale-if-error=86400'
```
rschristian, line 14: Bumped the cache age here as we're not releasing anywhere near often enough for an hour-long cache to really be justified.

JoviDeCroock (approved):

PR #1324, opened 2025-08-19
Bump to v11

JoviDeCroock:
Yikes, my npm version might be too new judging from the lockfile changes

### DIFF
--- a/src/components/header/index.jsx
+++ b/src/components/header/index.jsx
@@ -134,7 +134,7 @@ class NavMenu extends Component {
 	handleClickOutside = ({ target }) => {
 		if (this.state.open) {
 			do {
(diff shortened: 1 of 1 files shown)


rschristian (approved):
I'm on mobile so no lockfile diff, might be the first time it's being formatted (precommit hook)? I think I altered the filter on that recently, we might not've added or removed anything since.

JoviDeCroock:
Yes you are right, it's prettier diff, it does look like we might want to exclude the markdown content as it sometimes formats lines where we depends on indicator arrows in comments

rschristian (approved):
The arrow issue I was aware of, though expecting people to use `--no-verify` when correcting might be unreasonable.

PR #1328, opened 2025-08-26
ci: Try adding compressed-size-action for build stats

rschristian:
Might be useful for easier eyeballing of some PRs

### DIFF
--- a/.github/size-plugin.yml
+++ b/.github/size-plugin.yml
@@ -1,2 +1,0 @@
-size-files:
-- size-plugin.json
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "prismjs": "^1.30.0",
     "rollup": "^2.79.1",
     "sucrase": "^3.35.0",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -91,7 +91,7 @@
     "prismjs": "^1.30.0",
     "rollup": "^2.79.1",
     "sucrase": "^3.35.0",
-    "vite": "npm:rolldown-vite@^7.1.4",
+    "vite": "^7.1.7",
```
rschristian, line 94: `rolldown-vite` is has a good handful of known issues w/ `manualChunks`, which we make use of during prerendering. Given that the speed up is basically negligible anyhow, I think switching back is for the best.

.github/size-plugin.yml
```diff
@@ -1,2 +0,0 @@
-size-files:
-- size-plugin.json
```
rschristian, line 2: AFAIK this is unused? I think it comes from back when we were using preact-cli? Not quite sure tbh.

PR #1329, opened 2025-09-03
[fix]Support CRLF in frontmatter parsing

SisyphusZheng:
Updated the frontmatter regex to accept both LF and CRLF line endings, improving compatibility with Windows environments.


JoviDeCroock (approved):

rschristian:
Doesn't matter too much, but shouldn't all line endings be `lf`?

https://github.com/preactjs/preact-www/blob/ccde54fedbe7df0fbccd00d12abe159565fcfe0b/.editorconfig#L5

PR #1331, opened 2025-09-04
[fix]Add validation for Time component value prop

SisyphusZheng:
The Time component now checks if the 'value' prop is a non-empty string and logs warnings for invalid or unparsable values. This prevents rendering with invalid dates and improves error handling.

### DIFF
--- a/src/components/time/index.jsx
+++ b/src/components/time/index.jsx
@@ -1,13 +1,32 @@
 import s from './time.module.css';
 
 export function Time({ value }) {
-	const date = new Date(value + 'T00:00');
-	const day = date.toLocaleDateString
-		? date.toLocaleDateString()
-		: date.toDateString();
-	return (
-		<time class={s.time} dateTime={date.toISOString()}>
-			{day}
-		</time>
-	);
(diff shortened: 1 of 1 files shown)


rschristian:
Sorry, where did you run into this & what exactly does it solve? That component should only be used with our blog data which, as far as I can see, should all parse just fine?

If anything I'd rather have this throw & die as we should only be passing valid dates.

SisyphusZheng:
> Sorry, where did you run into this & what exactly does it solve? That component should only be used with our blog data which, as far as I can see, should all parse just fine?
> 
> If anything I'd rather have this throw & die as we should only be passing valid dates.

So here's the thing: I've forked preact-www in a clean Windows environment, and I've run npm install and npm run dev. My original intention was just to translate v11, but I've encountered several minor errors that are preventing some web page content from functioning. I'll provide the error logs I've encountered later.

rschristian:
Strange. I don't doubt you at all, though it's unclear to me what could fail there as it all seems right. Would love to know more.

However, we absolutely don't want to be validating our own data. It should be correct or fail to build outright.

PR #1340, opened 2025-09-21
Fix:Remove hidden visibility from current menu item

SisyphusZheng:
I personally believe that the current performance of the website on mobile devices is not optimal, and it would be better if improvements are made.

before:

before-1
before-2

after:
after-2

after-1

### DIFF
--- a/src/components/header/style.module.css
+++ b/src/components/header/style.module.css
@@ -141,10 +141,6 @@
 					padding: 0;
 				}
 
-				&.current {
-					visibility: hidden;
-				}
-
 				svg {
 					position: relative;
 					padding-right: 5px;


src/components/header/style.module.css
```diff
@@ -141,10 +141,6 @@
 					padding: 0;
 				}
 
-				&.current {
-					visibility: hidden;
-				}
-
```
rschristian, line 147: "Performance", or personal aesthetic preferences? There's a very large difference between the PR description and the actual changes, at least in English. Might just be a translation issue but wanted to ask in case I'm misunderstanding you.

Whilst I do prefer always showing the branding/home link on sites I create, I don't think we should be changing this just for personal preference. However, we have spoken in the past about heading styling issues on certain screen sizes (I think near 1000px wide is a bad one) and those I'd absolutely love fixes for, just don't have time myself.

SisyphusZheng: > "Performance", or personal aesthetic preferences? There's a very large difference between the PR description and the actual changes, at least in English. Might just be a translation issue but wanted to ask in case I'm misunderstanding you.
> 
> Whilst I do prefer always showing the branding/home link on sites I create, I don't think we should be changing this just for personal preference. However, we have spoken in the past about heading styling issues on certain screen sizes (I think near 1000px wide is a bad one) and those I'd absolutely love fixes for, just don't have time myself.

Oh,what I mean is that there is a gap in the "head" section when displayed on mobile devices(before-1-picture), which doesn't look very aesthetically pleasing. I just want to fix this issue, and displaying the logo directly is simply because this method is more straightforward.

PR #1352, opened 2025-10-17
Optimizing useComputed perf

JoviDeCroock:
Related to https://github.com/preactjs/signals/issues/772

### DIFF
--- a/content/en/guide/v10/signals.md
+++ b/content/en/guide/v10/signals.md
@@ -306,6 +306,35 @@ function Counter() {
 }
 ```
 
+> **Optimizing `useComputed` performance**
+>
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
@rschristian The decision wasn't made with eye on not being performant by default it was made in the essence of giving the user more freedom. You can optimize out of this situation but the alternative, where we only hold on to the initial compute function can't be optimized out of, in case the user leverages props or the identity of a signal changes that's something that is an inherent bug.

In eye of the above, do you think we should revisit the decision then?

developit:
another option we could consider is this signature:

```js
// runs only when touched signals change:
useComputed(() => { ... });

// runs when touched signals change OR any of the deps referenced change:
useComputed(() => { ... }, [a, b, c]);

// version that approximates current main behavior:
const callback = () => { ... };
useComputed(callback, [callback]);

// likely most common usage:
const preview = useComputed(() => text.value ?? props.text, [props.text]);
```

marvinhagemeister:
fair point too

PR #1354, opened 2025-10-19
refactor: Delegate all prefetching/preloading at app root

rschristian:
One handler at app root should replace all those individual instances nicely, resulting in wayyyy fewer handlers. 72 (combined) -> 2 `mouseover`/`touchstart` on the 'Getting Started' page.

### DIFF
--- a/src/lib/use-delegated-prefetch.js
+++ b/src/lib/use-delegated-prefetch.js
@@ -15,8 +15,6 @@ export function useDelegatedPrefetch() {
 				if (!e.target.href.startsWith(location.origin)) return;
 				const pathname = new URL(e.target.href).pathname;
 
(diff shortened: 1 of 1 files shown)


src/lib/use-delegated-prefetch.js
```diff
@@ -0,0 +1,42 @@
...
+} from '../components/routes.jsx';
+
+export function useDelegatedPrefetch() {
+	useEffect(() => {
+		const prefetchAndPreload = e => {
+			if (e.target.tagName === 'A') {
+				if (!e.target.href.startsWith(location.origin)) return;
+				const pathname = new URL(e.target.href).pathname;
+
+				if (!pathname.startsWith('/')) return;
```
rschristian:
```suggestion
```

PR #1355, opened 2025-10-22
Update index.jsx

aakanksha-grover:


rschristian (commented):
Do you work at Apple and have you gotten approval for this? We require folks only add logos with the company's permission.

PR #1363, opened 2025-12-06
chore: Add 'aria-label' for all nav areas

rschristian:
Fiddling with Algolia and went to exclude the ToC from indexing and realized we don't have any label for it at the moment.

Potentially could add a real label, though I think it's unnecessary. We should have an `aria-label` for it though at the least

Edit: On second thought, all of our nav areas should have aria labels. Will fix.

### DIFF
--- a/src/components/header/index.jsx
+++ b/src/components/header/index.jsx
@@ -51,7 +51,7 @@ function MainNav() {
 	};
 
 	return (
-		<nav class={style.nav}>
+		<nav class={style.nav} aria-label="Main navigation">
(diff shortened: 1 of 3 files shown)


rschristian:
^ We have two `style-<hash>.css` files output & compressed-size action is getting a bit confused and comparing one against the other. It doesn't make any real attempt at comparing sizes to find the likely comparison candidate, might be a good feature for the future.

No changes here are actually causing such a big difference.

JoviDeCroock (approved):

PR #1364, opened 2025-12-06
fix: Correct a few docsearch bugs

rschristian:
See comments

### DIFF
--- a/src/components/header/search.jsx
+++ b/src/components/header/search.jsx
@@ -29,10 +29,11 @@ function injectDocsearchCSS() {
  * interacted with the placeholder button prior to Docsearch fully loading.
  *
  * @param {HTMLElement} root
- * @returns {Promise<HTMLElement>}
+ * @returns {Promise<HTMLButtonElement>}
  */
 function waitForDocsearch(root) {
 	return new Promise(resolve => {
+		/** @returns {HTMLButtonElement | null} */
 		const getDocSearchButton = () =>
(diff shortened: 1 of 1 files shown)


src/components/header/search.jsx
```diff
@@ -69,10 +70,6 @@ export default function Search() {
 	const root = useRef(null);
 	const rendered = useRef(false);
 	const interactedWith = useRef(false);
-	const loadingBar =
-		typeof window !== 'undefined'
-			? document.querySelector('loading-bar')
-			: null;
```
rschristian: Shouldn't be problematic once deployed, due to prerendering, but annoying in dev. The `<loading-bar>` component renders in the page, so after the header, and won't exist by the time it's captured here.

Very superficial, but silly left over that should've been caught.

src/components/header/search.jsx
```diff
@@ -107,26 +104,37 @@ export default function Search() {
...
+			{rendered.current ? (
+				<ErrorBoundary>
+					<DocSearch
+						apiKey={config.docsearch.apiKey}
+						indexName={config.docsearch.indexName}
+						appId={config.docsearch.appId}
+						transformItems={transformItems}
+					/>
+				</ErrorBoundary>
+			) : (
```
rschristian: After a user interacts with the search, such as clicking a link, it'll no longer function properly and the modal will not pop up as navigating rerenders the header and returns the placeholder content, wiping out the newly-rendered `<DocSearch>`.

PR #1367, opened 2025-12-29
doc: Add HesapKurdu to the list of companies in the We Are Using section

mert-solak:


rschristian:
Do you work at the company? We always have to ask.

PR #1394, opened 2026-08-26
docs: clarify keyed Fragment reconciliation

MuslumYilmaz:
## Summary

- Explain that a key identifies the whole Fragment during reconciliation.
- Clarify why stable keys matter when Fragment groups are reordered.
- Note that keyed Fragments require `<Fragment key={...}>`, since shorthand syntax cannot accept a key.
- Keep the v10 and v11 English guides aligned.

The current example correctly recommends keys but stops at "Preact has to guess," leaving the group-identity and reordering consequence implicit.

Docs-only change; the same clarification is applied to the English v10 and v11 guides so their Fragment sections remain aligned.

### DIFF
--- a/content/en/guide/v10/components.md
+++ b/content/en/guide/v10/components.md
@@ -203,13 +203,8 @@ function Glossary(props) {
 }
 ```
 
-
 A key identifies the entire Fragment as one group during reconciliation. When
 Fragment groups are reordered, stable keys help Preact match each group with
 the corresponding group from the previous render instead of matching only by
 position. Use the explicit `<Fragment key={...}>` form for keyed Fragments,
 because the `<>...</>` shorthand cannot accept a key.
-
-For a React-focused comparison and practice exercise covering wrapper-free DOM
-output and key-driven identity, see
-[Fragments, DOM, and reconciliation](https://frontendatlas.com/react/trivia/react-fragments-dom-and-reconciliation).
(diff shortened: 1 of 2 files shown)


content/en/guide/v10/components.md
```diff
@@ -202,3 +202,14 @@ function Glossary(props) {
...
 ```
+
+
+A key identifies the entire Fragment as one group during reconciliation. When
+Fragment groups are reordered, stable keys help Preact match each group with
+the corresponding group from the previous render instead of matching only by
+position. Use the explicit `<Fragment key={...}>` form for keyed Fragments,
+because the `<>...</>` shorthand cannot accept a key.
+
+For a React-focused comparison and practice exercise covering wrapper-free DOM
```
rschristian: Hard no on recommending any unknown & untrusted resources, sorry. At the very least it creates churn for us if you take down or move the site at any point. Hosting content ourselves is better wherever possible.

If you think our current tutorial lesson is lacking, we can expand it

rschristian (commented):
Content is otherwise good

content/en/guide/v10/components.md
MuslumYilmaz: Understood, thanks for clarifying. I’ve removed the third-party link from both versions and kept the Fragment clarification self-contained. I’ll leave any tutorial expansion for a separate contribution.

rschristian (approved):
Thanks!

PR #1401, opened 2026-09-13
chore: Bump more simple deps

rschristian:


src/components/controllers/repl/repl.worker.js
```diff
@@ -1,4 +1,3 @@
-import '@babel/polyfill';
```
rschristian, line 1: This was added to support `sucrase` in older browsers in #422

However, per [this comment](https://github.com/preactjs/preact-www/pull/422#issuecomment-531367892), it didn't really achieve the goal back then and all of those browsers are now very ancient today. We can drop the polyfill (which also emits a handful of warnings when installing).

rschristian:
Something seems broken w/ the way we're lazy loading the algolia search widget now, putting into draft until I can figure out if that's just in the preview env

PR #1404, opened 2026-09-14
refactor: Add title to blog index page & simplify content

rschristian:


content/en/blog.md
```diff
@@ -1,18 +1,18 @@
 ---
 title: Blog
+description: Find out more about what the Preact team is working on
 ---
 
+# Preact Team Blog
```
rschristian, line 6: I've gone back and forth on this, both in form (it existing at all) and actual text.

On one hand, every page should really have an h1, currently we have only h2 values. Whilst we could make it hidden (`position: absolute; width: 1px; height: 1px:`, etc), arguably it should be on the page. Just "Blog" looked a bit... redundant, and plain, so I fiddled with it a bit and found "Preact Team Blog" filled the page out a bit more.

Looking for thoughts & opinions though.

src/components/blog-overview/index.jsx
```diff
@@ -8,27 +8,25 @@ export default function BlogOverview() {
 	const translate = useTranslate();
 
 	return (
-		<div>
-			<div class={s.postList}>
-				{Object.entries(blogPosts).map(([postPath, post]) => {
-					const translatedBlog = translate('blogPosts', post.label);
+		<div class={s.postList}>
+			{Object.entries(blogPosts).map(([postPath, post]) => {
+				const translatedBlog = translate('blogPosts', post.label);
```
rschristian, line 13: Diff here is just shedding the extraneous wrapping `<div>`, it probably had a class or something at one point but it's redundant today.

JoviDeCroock (approved):

PR #1406, opened 2026-09-17
Add Extension.js to Integrations

cezaraugusto:
[Extension.js](https://github.com/extension-js/extension.js) builds browser extensions with Preact and no build config. one manifest.json produces the Chrome, Edge, Firefox and Safari build, `extension dev` opens the browser with the extension loaded, and popup, options and devtools pages hot reload their Preact components

sample:

https://github.com/user-attachments/assets/650249c9-8bc3-4012-a81c-3de4df442774

docs: https://extension.js.org, and the Preact templates open in the browser at https://templates.extension.dev/preact

I maintain it, MIT


rschristian (approved):
Thanks!

